PR #7, opened 2014-08-28
GRADLE-3158 Cannot copy between different mount points with Gradle 2.0 o...

SchwingSK:
...n Ubuntu

- use g++ compiler option '-D_FILE_OFFSET_BITS=64' to handle stat responses from Windows shares (64bit Inode values)


sschuberth:
@adammurdoch This has been pending since 2014, mind having a look again?

PR #17, opened 2016-10-03
Add z System (s390x) support

nealef:
Hi,
 I have some small patches that enable native-platform to build and run on IBM z Systems (aka s390x) under Linux. I couldn't find a mailing list to open discussions or find out what the process for adding a platform is, so I am just creating this PR to introduce the addition. I can provide the build log to show the build/test process is clean if that would be helpful. 

Regards, Neale Ferguson

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -107,6 +107,14 @@ model {
             architecture "amd64"
             operatingSystem "linux"
         }
+        linux_s390x {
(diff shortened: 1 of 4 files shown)


ov7a:
Related discussion on adding more platforms will happen in the following PR
* https://github.com/gradle/native-platform/pull/361

PR #19, opened 2016-11-13
Prefer ncurses 6 over 5

szpak:
In most cases it should not make a difference (having both ncurses 6 and legacy 5 installed), however version 6 is newer and in a long run has a chance to contain some bug fixes and enhancements not available in version 5.


adammurdoch:
Thanks for the pull request.

PR #21, opened 2016-12-20
MemoryInfo implementation for OSX

eskatos:
MemoryInfo should now be implemented for OS X.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,7 +19,7 @@ allprojects {
     }
 
     group = 'net.rubygrapefruit'
-    version = '0.13'
+    version = '0.14'
 
     if (!project.hasProperty('release')) {
         version = "${version}-dev"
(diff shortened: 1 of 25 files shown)


eskatos:
The added method is in `src/shared/cpp/osx.cpp`, shouldn't it be in `src/main/cpp/apple.cpp` instead?

I didn't use `vm_stats.external_page_count`, which is shown as `File-backed pages` by the `vm_stat`command, because old versions of `host_statistics64` do not return it. Plus, my laptop reported way too big numbers for this value after being on for weeks, it got back to a sensible value after rebooting.

With regards to testing, shall we restrict the test to run on OS X only?
I also have local changes to assert that total memory is correct by comparing it to what JMX reports, and that available memory is bigger than what JMX reports. Sounds good?

eskatos:
Please ignore this PR, I'll reopen another one shortly.

PR #22, opened 2017-01-05
MemoryInfo implementation for MAC

eskatos:
Uses `sysctl` to query total physical memory.

Uses `host_statistics64` to calculate available physical memory (`free_count + inactive_count - speculative_count`).

Do not rely on `external_page_count` which should be the file cache because it is available since OSX 10.9 only and it shown inaccurate (larger than total memory) values with long uptime or under load.

Subtract `speculative` in order to provide a pessimistic result rather than an optimistic one.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -68,7 +68,8 @@ task nativeHeaders {
             'net.rubygrapefruit.platform.internal.jni.WindowsFileFunctions',
             'net.rubygrapefruit.platform.internal.jni.FileEventFunctions',
             'net.rubygrapefruit.platform.internal.jni.PosixTypeFunctions',
-            'net.rubygrapefruit.platform.internal.jni.MemoryFunctions'
+            'net.rubygrapefruit.platform.internal.jni.MemoryFunctions',
+            'net.rubygrapefruit.platform.internal.jni.OsxMemoryFunctions'
     ]
     inputs.files sourceSets.main.output
     inputs.property('classes', classes)
--- a/build.gradle
+++ b/build.gradle
@@ -67,7 +67,8 @@ task nativeHeaders {
             'net.rubygrapefruit.platform.internal.jni.WindowsRegistryFunctions',
             'net.rubygrapefruit.platform.internal.jni.WindowsFileFunctions',
             'net.rubygrapefruit.platform.internal.jni.FileEventFunctions',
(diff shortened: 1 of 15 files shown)


eskatos:
Here are some evidence supporting the free memory calculation:
- https://developer.apple.com/library/content/documentation/Performance/Conceptual/ManagingMemory/Articles/AboutMemory.html
- https://github.com/oshi/oshi/blob/master/oshi-core/src/main/java/oshi/hardware/platform/mac/MacGlobalMemory.java

eskatos:
Here's a sh file to quickly do `vm_stat` queries: https://gist.github.com/eskatos/8c05edb1c2ae0bd20a490fefb9724c66

eskatos:
I dug into OSX `vm_stat` source code from 10.0 to 10.12 and here's what i've learnt.

The source is available here https://opensource.apple.com/ in the `system_cmds` archives.

**10.0 to 10.5** (>2001)
Uses `host_statistics`
`Free pages` is `stats.free_pages`
`Speculative pages` is not present
`File-backed pages` is not present

**10.6 to 10.8** (>2009)
Uses `host_statistics64`
`Free pages` is `stats.free_pages - stats.speculative_pages`
`File-backed pages` is not present

**10.9 to 10.X** (>2013)
Uses `host_statistics64`
`Free pages` is `stats.free_pages - stats.speculative_pages`
`File-backed pages` is `stats.external_page_count`

eskatos:
wrt ☝️ 
I don't think we care so much about <=10.5
What we do today in Gradle, parsing `vm_stat` output and using `File-backed pages` do not work prior to 10.9

eskatos:
I've added `OsxMemoryInfo` as an experiment to get detailled OSX memory details and tests to demonstrate the deviation from `vm_stat` values I've been observing.

There are two tests in `OsxMemoryTest`. The first one exercise native memory info gathering and outputs a CSV with values from both native `host_statistics64` and parsing the `vm_stat` output. The second test does the same repeatedly and indefinitely, and puts the CSVs in `/tmp`. The latter is `@Ignore`d to not block the build but easy to use and run with various delays to collect some data.

Here is some sample data: https://gist.github.com/eskatos/86ddcac0e6b3e92cb11ca329a4198370

eskatos:
@adammurdoch this should be in a pretty good shape and the calculation is as we discussed it

PR #30, opened 2018-12-18
Allow stat() to handle paths that happen to match an existing file 

ghale:
There are a couple of things in this pull request:

- Changes so that `stat()` does not return an exception when something in the path happens to match some existing file.  It now treats it like the path is not found rather than an error.
- Changes to build with Xcode10 on OSX.  This also drops support for OSX 32-bit.
- Some updates/refactors to tests to get them to pass on various platforms:
  - Ignore the test for file names longer than the 260 character max on windows.
  - Resolve the timestamp difference between `stat()` and java file methods on macOS.
  - Fix an issue on OSX/jdk8 where closing the file watch was causing an EBADF instead of EINTR when polling for the next change.

The tests all pass for me now on both jdk7 and jdk8 on OSX, Linux and Windows.  I did not find any jvm-specific differences that seemed to require any special handling (at least on the OS/JVM combinations I tested with).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -99,10 +99,6 @@ def ncursesVersion = inferNCursesVersion(os)
 
 model {
     platforms {
-        osx_i386 {
-            architecture "i386"
-            operatingSystem "osx"
-        }
         osx_amd64 {
             architecture "amd64"
             operatingSystem "osx"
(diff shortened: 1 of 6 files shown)


adammurdoch:
Looks good, thanks.

PR #32, opened 2019-02-07
Resolve issues with supplementary characters being set in environment variables

ghale:
This mitigates the issue by always using utf-8 encoding when getting/setting environment variables.  I feel like this is the simplest solution to the problem.  Other places where we convert, we continue to use the C-locale-dependent conversion methods, but for environment variables, we always use utf-8.

The issue with converting to bytes on the Java side and passing a byte array in is that on Windows, we don't set the environment variable using `char *`, we need `wchar *`, so if we pass in bytes then we have the opposite problem and we have to do the same conversion on windows now and potentially open ourselves to the same problem.

I also looked into doing a C-locale-dependent conversion first, and then trying the Java-byte-conversion method, but I found scenarios where this would fail.   Basically, the `setEnvironmentVariable` would fail on the C-locale-dependent conversion, we'd fallback to Java-byte-conversion, which would succeed in putting it into the environment, but then on the call to `getEnvionmentVariable` we would not fail on the C-locale-dependent conversion, it would succeed, only the result would be garbage.  It seemed safer to ensure we always convert the same way on both set/get.

There are certainly things we could do to make the Java-byte-conversion method work without needing to mess with the Windows methods, but just using the built-in utf8 conversion (for environment variables only) seemed like the safest, simplest approach. 

Let me know what you think.

### DIFF
--- a/src/main/cpp/posix.cpp
+++ b/src/main/cpp/posix.cpp
@@ -328,16 +328,20 @@ Java_net_rubygrapefruit_platform_internal_jni_PosixProcessFunctions_getEnvironme
 JNIEXPORT void JNICALL
 Java_net_rubygrapefruit_platform_internal_jni_PosixProcessFunctions_setEnvironmentVariable(JNIEnv *env, jclass target, jstring var, jstring value, jobject result) {
     char* varStr = java_to_char(env, var, result);
-    if (value == NULL) {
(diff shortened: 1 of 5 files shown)


adammurdoch:
Merged, thanks.

PR #34, opened 2019-05-13
Windows: Add support for paths longer than 260 characters

rpaquay:
* For local paths, we use the `\\?\` prefix
* For UNC paths, we use the `\\?\UNC\` prefix
* Added a `--long-paths` option to the test app to verify behavior

### DIFF
--- a/test-app/src/main/java/net/rubygrapefruit/platform/test/Main.java
+++ b/test-app/src/main/java/net/rubygrapefruit/platform/test/Main.java
@@ -42,7 +42,6 @@ public class Main {
         optionParser.accepts("ansi", "Force the use of ANSI escape sequences for terminal output");
         optionParser.accepts("stat", "Display details about the specified file or directory").withRequiredArg();
         optionParser.accepts("ls", "Display contents of the specified directory").withRequiredArg();
-        optionParser.accepts("long-paths", "Test support for long (i.e. >= 260 characters) paths");
         optionParser.accepts("watch", "Watches for changes to the specified file or directory").withRequiredArg();
         optionParser.accepts("machine", "Display details about the current machine");
         optionParser.accepts("terminal", "Display details about the terminal");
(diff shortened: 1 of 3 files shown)


adammurdoch:
Thanks for the PR. I'll review it soon

test-app/src/main/java/net/rubygrapefruit/platform/test/Main.java
```diff
@@ -388,6 +394,74 @@ private static void ls(String path) {
         }
     }
 
+    private static void longPaths() {
```
adammurdoch: I think this would be better as a unit test, for example in `FilesTest` or perhaps a new `WindowsFilesTest` class. This way we can check that it still works every time we run the tests (eg when releasing).

rpaquay: Done. The FilesTest file already had test for this (they were simply skipped on Windows), so I just had to remove those checks and verify the test pass (they do).

I also removed the "longPaths" option from the test app since it is not needed.

adammurdoch:
Great, thank you for making the changes. I'll merge this PR soon and review your other PRs also.

lacasseio:
A general comment, I think we could add tests for those long paths explicitly.

rpaquay:
@lacasseio I am a little confused by your comment. The unit tests had explicit long paths tests that were disabled on Windows (see https://github.com/adammurdoch/native-platform/commit/3c1296467b2e2bd5e0f6ff5a537e9e3721283576), so re-enabling them seems like the natural thing to do when adding support for long paths on that platform.

lacasseio:
@rpaquay You are right, my bad for the noise, I should have looked more closely.

rpaquay:
@adammurdoch is there still something you are expecting me to change for this particular PR? (I will rebase the others PRs once this one is merged to make them easier to review).

adammurdoch:
@rpaquay No changes needed. I just need to find some time to test out the changes and merge them. I'll try to do this soon.

adammurdoch:
Merged, thanks for the PR.

rpaquay:
Thanks!

PR #38, opened 2019-05-22
Windows: Add support for returning file IDs from `stat` and `readdir`

rpaquay:
* Add `FileInfo.getKey()` that returns a file identifier than can be used to compare
  efficiently files for equality. If the key is not available, the method returns
  `null`.
* This is a similar concept as exposed in the Java 7 nio API
  https://docs.oracle.com/javase/7/docs/api/java/nio/file/attribute/BasicFileAttributes.html#fileKey()
* A WindowsFileKey is a pair of (volume serial number, file id on the volume)
* Update the `stat` and `fastReaddirXxx` jni entry points to provide this information
* Update the test app to display `file key` if it is available


lacasseio:
Make sure only the commit for this PR is included.

PR #42, opened 2019-10-30
Add system hostname

tomdcc:
Currently for posix systems this uses the nodename returned by uname rather than the hostname call as a) we are already calling uname and b) on MacOS and Linux these apparently return the same thing anyway.

Windows implementation will probably use GetComputerName, yet to be implemented.


adammurdoch:
Merged, thanks for the PR.

PR #44, opened 2019-11-26
Allow publishing partial snapshot variant

wolfs:
This PR allows publishing snapshots, which use a timestamped version number. Moreover, it is possible to publish snapshots with fewer supported OSes to the Gradle snapshot repository.

### DIFF
--- a/buildSrc/src/main/java/UploadPlugin.java
+++ b/buildSrc/src/main/java/UploadPlugin.java
@@ -2,6 +2,7 @@ import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.artifacts.repositories.MavenArtifactRepository;
+import org.gradle.api.publish.Publication;
 import org.gradle.api.publish.PublishingExtension;
 import org.gradle.api.publish.maven.MavenPublication;
 import org.gradle.internal.ErroringAction;
@@ -33,19 +34,12 @@ public class UploadPlugin implements Plugin<Project> {
                 UpdatePackageMetadataTask update = project.getTasks().create("updatePackage" + capitalizedPublicationName, UpdatePackageMetadataTask.class);
                 update.setPublication(publication);
 
-                UploadTask upload = project.getTasks().create("uploadPackage" + capitalizedPublicationName, UploadTask.class);
+                UploadTask upload = project.getTasks().create(uploadTaskName(publication), UploadTask.class);
                 upload.setGroup("Upload");
                 upload.setDescription("Upload publication " + publication.getName());
                 upload.setLocalRepoDir(repoDir);
                 upload.setPublication(publication);
                 upload.dependsOn(update);
-                if (!BasePublishPlugin.isMainPublication(publication)) {
-                    Task uploadLifecycle = project.getTasks().maybeCreate("uploadJni");
-                    uploadLifecycle.setGroup("Upload");
-                    uploadLifecycle.setDescription("Upload all JNI publications");
-                    uploadLifecycle.dependsOn(upload);
-                }
-
             });
         });
         project.getGradle().getTaskGraph().whenReady(graph -> {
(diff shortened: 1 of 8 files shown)


wolfs:
This seems to work now. See this CI job and its dependencies: https://builds.gradle.org/viewLog.html?buildId=29004031&buildTypeId=GradleNative_NativePlatform_Publishing_PublishJavaApiSnapshot

I managed to publish the snapshot version `0.20-snapshot-20191126174414+0000` from there: https://repo.gradle.org/gradle/webapp/#/artifacts/browse/tree/General/ext-snapshots-local/net/rubygrapefruit/native-platform/0.20-snapshot-20191126174414+0000

@adammurdoch Can you take a look?

buildSrc/src/main/java/UploadPlugin.java
```diff
@@ -16,64 +14,47 @@
...
-                            uploadLifecycle.dependsOn(upload);
-                        }
+                UploadTask upload = project.getTasks().create("uploadPackage" + capitalizedPublicationName, UploadTask.class);
+                upload.setGroup("Upload");
+                upload.setDescription("Upload publication " + publication.getName());
+                upload.setLocalRepoDir(repoDir);
+                upload.setPublication(publication);
+                upload.dependsOn(update);
+                if (!BasePublishPlugin.isMainPublication(publication)) {
+                    Task uploadLifecycle = project.getTasks().maybeCreate("uploadJni");
```
adammurdoch: Maybe give `uploadJni` and `publishJni` more descriptive names. For a while I didn't notice that they were different.

Alternatively, which was the original intent for this task, let's have a single task called `publishJni` whose role is to publish the JNI libraries to the correct repository for the kind of output the build is producing. That is, that publishes the JNI libraries to the snapshot repo if we're building a snapshot, or bintray if we're building a release. It would also be good to have a similar lifecycle task for the main publications too.

wolfs: This is the current behaviour:

There are currently two types of tasks used for publishing to a remote repository:
- `uploadPackage*`, generated by the `UploadPlugin`, which use the Bintray API to upload/publish things to Bintray. Those should be used for final releases
- `publish*PublicationToSnapshotRepository`, created by the `maven-publish` plugin

For the `uploadPackage*` tasks there is an `uploadJni` lifecycle task which uploads all the native artifacts to Bintray. Moreover, there is an `uploadPackageMain`, which uploads the main JAR to Bintray. The upload tasks require publishing the artifact to a local Maven repository first, this is done by `publish*PublicationToMavenRepository`.

For the `publish*PublicationToSnapshotRepository`, there is a `publishJni` lifecycle tasks, which publishes all the native artifacts to the snapshot repository. Moreover, there is the `publishMainPublicationToSnapshotRepository` task which does the same for the main JAR.

Currently, only snapshot releases are published to the snapshot repository (`ext-snapshots-local`), though currently nothing stops you from publishing other releases there as well.

Releases and milestones are currently published to Bintray, though nobody would stop you currently to upload snapshots to Bintray as well.

wolfs: So there are a few ways forward, to make this less confusing:
1. rename `publishJni` to `publishJniPublicationsToSnapshotRepository`, so it is more in line with the other publish tasks.
2. use lifecycle tasks (`uploadJni`/`uploadMain` maybe?), to upload the artifacts to the correct place depending on the kind of release.

It seems like the second option would be better. WDYT?

wolfs: I implemented 2). @adammurdoch PTAL!

adammurdoch: I agree. I'd prefer to have a lifecycle task that just does the right thing for the type of build that's running (option 2), rather than having to both set the project property and run specific upload tasks and remember to use the correct combination of these things (option 1).

adammurdoch: btw, you should be able to merge this PR. Feel free to merge it and then make improvements in a follow up PR.

PR #47, opened 2019-11-29
MacOs file watcher

gavra0:
Work in progress

Initial implementation for MacOs file watcher.


src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -40,9 +39,7 @@ static void reportEvent(const char *event, char *path) {
     JNIEnv* env;
     int getEnvStat = jvm->GetEnv((void **)&env, JNI_VERSION_1_6);
     if (getEnvStat == JNI_EDETACHED) {
-        printf("~~~~ GetEnv: not attached");
         if (jvm->AttachCurrentThread((void **) &env, NULL) != 0) {
-            printf("~~~~ Failed to attach");
```
lptr: Let's add `// TODO`-style comments here to make sure we don't forget error handling.

src/main/java/net/rubygrapefruit/platform/internal/jni/DefaultOsxFileEventFunctions.java
```diff
@@ -35,7 +35,9 @@ public void pathChanged(String path) {
         }
 
         private List<String> getAllChanges() {
-            return new ArrayList<String>(globalQueue);
+            ArrayList<String> result =  new ArrayList<String>(globalQueue);
+            globalQueue.clear();
```
lptr: Oh, right, that was me. Sorry!

src/main/cpp/apple_fsnotifier.cpp
gavra0: I've added error handling using `FunctionResult` object.

src/main/java/net/rubygrapefruit/platform/internal/jni/DefaultOsxFileEventFunctions.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Queue;
+import java.util.Set;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+public class DefaultOsxFileEventFunctions implements NativeIntegration {
+
+    public boolean addRecursiveWatch(String path) {
```
adammurdoch: We should add some synchronization and lifecycling to prevent things like adding watches from multiple threads at the same time, or adding watches while we're already watching, or starting watching while we're already watching, and so on.

src/main/java/net/rubygrapefruit/platform/internal/jni/DefaultOsxFileEventFunctions.java
```diff
@@ -0,0 +1,73 @@
...
+        ChangeCollector collector = new ChangeCollector();
+        FunctionResult result = new FunctionResult();
+        OsxFileEventFunctions.startWatch(collector, result);
+        if (result.isFailed()) {
+            throw new NativeException("Failed to start collecting changes. Reason: " + result.getMessage());
+        }
+        return collector;
+    }
+
+    public List<String> stopWatch(ChangeCollector collector) {
```
adammurdoch: There's so much static state in the implementation that it seems odd that the collector is a parameter here, rather than just more static state.

Same for `startWatch()` and its return value.

src/test/groovy/net/rubygrapefruit/platform/file/OsxFileEventsTest.groovy
```diff
@@ -0,0 +1,155 @@
...
+
+        when:
+        def changes = fileEvents.stopWatch(collector);
+
+        then:
+        changes == [dirA.canonicalPath + "/", dirB.canonicalPath + "/"]
+    }
+
+    // TODO: this is not great, as it leads to flaky tests. Figure out a better way.
+    private static void waitForFileSystem() {
```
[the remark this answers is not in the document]
lptr: Yes, we definitely should do this.

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -0,0 +1,179 @@
...
+        invalidStateDetected = false;
+        rootsToWatch = CFArrayCreateMutable(NULL, 0, NULL);
+        if (rootsToWatch == NULL) {
+            mark_failed_with_errno(env, "Could not allocate array to store roots to watch.", result);
+            return;
+        }
+    }
+    char* pathString = java_to_char(env, path, result);
+    if (pathString == NULL) {
+        return;
```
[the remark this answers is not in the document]
lptr: I'd like to revamp our `null` handling completely when we move to the even- driven API. Let's keep it this way for now as it will be refactored later.

src/main/java/net/rubygrapefruit/platform/internal/jni/DefaultOsxFileEventFunctions.java
```diff
@@ -0,0 +1,73 @@
...
+import net.rubygrapefruit.platform.internal.FunctionResult;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Queue;
+import java.util.Set;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+public class DefaultOsxFileEventFunctions implements NativeIntegration {
```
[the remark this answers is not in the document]
lptr: 👍

src/main/java/net/rubygrapefruit/platform/internal/jni/DefaultOsxFileEventFunctions.java
lptr: 👍

lptr:
I'll merge this as is, and will follow up on the notes in a following PR.

PR #48, opened 2019-12-02
Remove old file events

lptr:
We are introducing a new API for specific OSs instead. The old one never really worked and should never have been used anywhere.


test-app/src/main/java/net/rubygrapefruit/platform/test/Main.java
```diff
@@ -353,41 +347,6 @@ private static void fileSystems() {
         System.out.println();
     }
 
-    private static void watch(String path) {
```
adammurdoch: It would be nice to keep this in some form and change it to use the new API

lptr: How is this used? We can add this back once we have working support for some of the platforms.

adammurdoch: It's part of a test/showcase app that you can use for manually trying out the APIs

PR #50, opened 2019-12-05
Nicer file watcher API (for macOS for now)

lptr:
This is about making the API event-driven, so we can get notifications on the Java side as soon as possible, and in general making it nicer to work with.

The idea is that we could extend this API to cover Windows (and potentially Linux as well).

/cc @gavra0 @wolfs


lptr:
I'd also like to explore moving all the static state into a `Closable` object returned by `startWatch()`.

src/main/java/net/rubygrapefruit/platform/file/FileWatcher.java
```diff
@@ -0,0 +1,20 @@
+package net.rubygrapefruit.platform.file;
+
+import java.io.Closeable;
+import java.io.IOException;
+
+/**
+ * A handle for watching file system locations.
+ */
+public interface FileWatcher extends Closeable {
```
wolfs: Shouldn't this be called `FileWatch`?

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
```diff
@@ -16,10 +16,51 @@
...
+        return watch;
+    }
+
+    private static native FileWatcher startWatch(String[] path, double latency, FileWatcherCallback callback, FunctionResult result);
+    private static native void stopWatch(Object details, FunctionResult result);
+
+    // Created from native code
+    @SuppressWarnings("unused")
+    private static class WatchImpl implements FileWatcher {
+        private Object details;
```
wolfs: Can you add some doc to the `details` object what it is used for? I took me some time to understand. Declaring the type to `java.nio.ByteBuffer` would help as well.

src/test/groovy/net/rubygrapefruit/platform/file/OsxFileEventsTest.groovy
```diff
@@ -30,122 +30,138 @@ import spock.lang.Timeout
 class OsxFileEventsTest extends Specification {
```
wolfs: ☝️ How about `@Requires({ Platform.current().macOs })` instead of `@IgnoreIf`?

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
```diff
@@ -16,10 +16,51 @@
...
 
-public class OsxFileEventFunctions {
-    public static native void createWatch(String path, FunctionResult result);
-    public static native void startWatch(DefaultOsxFileEventFunctions.ChangeCallback callback, FunctionResult result);
-    public static native void stopWatch(FunctionResult result);
+import java.util.Collection;
+
+public class OsxFileEventFunctions implements NativeIntegration {
+
+    public FileWatcher startWatch(Collection<String> paths, double latency, FileWatcherCallback callback) {
```
wolfs: How about `startWatching`?

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
```diff
@@ -16,10 +16,16 @@
...
 package net.rubygrapefruit.platform.internal.jni;
 
+import net.rubygrapefruit.platform.file.FileWatcher;
 import net.rubygrapefruit.platform.internal.FunctionResult;
 
 public class OsxFileEventFunctions {
-    public static native void createWatch(String path, FunctionResult result);
-    public static native void startWatch(DefaultOsxFileEventFunctions.ChangeCallback callback, FunctionResult result);
-    public static native void stopWatch(FunctionResult result);
+    public static native FileWatcher startWatch(String[] path, double latency, ChangeCallback callback, FunctionResult result);
```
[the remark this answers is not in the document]
lptr: This method is not part of the public API, and is only use inside this class with an actual array, and thus there's no need for supporting varargs.

src/main/java/net/rubygrapefruit/platform/file/FileWatcher.java
lptr: I don't feel strongly either way. Arguments?

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
lptr: The `details` are just whatever the native code wants to keep track of. I will add docs.

I want to keep this as an opaque `Object`, though. The Java side has no business with the contents, so it shouldn't have any knowledge about it. `Object` captures that best I think, at least better than `ByteBuffer`.

lptr: FTR, I want to reuse this type for Windows as well.

src/test/groovy/net/rubygrapefruit/platform/file/OsxFileEventsTest.groovy
lptr: Fixed.

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
lptr: OK, renamed.

src/main/java/net/rubygrapefruit/platform/file/FileWatcher.java
lptr: I'd leave it as `FileWatcher` actually. "Watch" as a noun reminds me of the Night's Watch in GoT.

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
```diff
@@ -16,10 +16,55 @@
...
+import java.util.Collection;
+
+public class OsxFileEventFunctions implements NativeIntegration {
+
+    public FileWatcher startWatching(Collection<String> paths, double latency, FileWatcherCallback callback) {
+        if (paths.isEmpty()) {
+            return FileWatcher.EMPTY;
+        }
+        FunctionResult result = new FunctionResult();
+        FileWatcher watch = startWatching(paths.toArray(new String[0]), latency, callback, result);
```
wolfs:
```suggestion
        FileWatcher watcher = startWatching(paths.toArray(new String[0]), latency, callback, result);
```

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
```diff
@@ -16,10 +16,55 @@
...
+        if (result.isFailed()) {
+            throw new NativeException("Failed to start watch. Reason: " + result.getMessage());
```
wolfs:
```suggestion
            throw new NativeException("Failed to start watching. Reason: " + result.getMessage());
```

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
```diff
@@ -16,10 +16,55 @@
...
+        }
+        return watch;
+    }
+
+    private static native FileWatcher startWatching(String[] paths, double latency, FileWatcherCallback callback, FunctionResult result);
+    private static native void stopWatch(Object details, FunctionResult result);
```
wolfs:
```suggestion
    private static native void stopWatching(Object details, FunctionResult result);
```

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
wolfs: 👍

src/test/groovy/net/rubygrapefruit/platform/file/OsxFileEventsTest.groovy
```diff
@@ -18,134 +18,150 @@ package net.rubygrapefruit.platform.file
...
-        waitForFileSystem()
-
-        def collector = fileEvents.startWatch()
-        new File(dirA, "a.txt").createNewFile();
-        new File(dirB, "b.txt").createNewFile();
-        waitForFileSystem()
-
-        when:
-        def changes = fileEvents.stopWatch(collector);
+    private List<String> startWatch(double latency = 0.3, String... paths) {
```
wolfs:
```suggestion
    private List<String> startWatching(double latency = 0.3, String... paths) {
```

src/test/groovy/net/rubygrapefruit/platform/file/OsxFileEventsTest.groovy
```diff
@@ -18,134 +18,150 @@ package net.rubygrapefruit.platform.file
...
+        watch = fileEvents.startWatching(paths as List, latency) {
+            println "> $it"
+            changes.add(it)
+        }
+        return changes
+    }
 
-        then:
-        changes == [dirA.canonicalPath + "/", dirB.canonicalPath + "/"]
+    private void stopWatch() {
```
wolfs:
```suggestion
    private void stopWatching() {
```

src/test/groovy/net/rubygrapefruit/platform/file/OsxFileEventsTest.groovy
```diff
@@ -18,134 +18,150 @@ package net.rubygrapefruit.platform.file
...
 
-@IgnoreIf({ !Platform.current().macOs })
+@Requires({ Platform.current().macOs })
 @Timeout(20)
 class OsxFileEventsTest extends Specification {
     @Rule
     TemporaryFolder tmpDir
-    final DefaultOsxFileEventFunctions fileEvents = Native.get(DefaultOsxFileEventFunctions.class)
+    final OsxFileEventFunctions fileEvents = Native.get(OsxFileEventFunctions.class)
+    FileWatcher watch
```
wolfs:
```suggestion
    FileWatcher watcher
```

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
```diff
@@ -16,10 +16,55 @@
...
+        @Override
+        public void close() {
+            if (details == null) {
+                return;
+            }
+            FunctionResult result = new FunctionResult();
+            stopWatch(details, result);
+            details = null;
+            if (result.isFailed()) {
+                throw new NativeException("Failed to stop watch. Reason: " + result.getMessage());
```
wolfs:
```suggestion
                throw new NativeException("Failed to stop watching. Reason: " + result.getMessage());
```

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -59,93 +61,113 @@ static void callback(ConstFSEventStreamRef streamRef,
...
 
+    // TODO Should this be somewhere global?
     int jvmStatus = env->GetJavaVM(&jvm);
     if (jvmStatus < 0) {
         mark_failed_with_errno(env, "Could not store jvm instance.", result);
-        return;
+        return NULL;
     }
+
+    jclass clsWatch = env->FindClass("net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions$WatcherImpl");
```
```suggestion
    jclass clsWatcher = env->FindClass("net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions$WatcherImpl");
```

src/test/groovy/net/rubygrapefruit/platform/file/OsxFileEventsTest.groovy
```diff
@@ -23,149 +23,164 @@ import org.junit.Rule
...
     }
 
     def "can open and close watcher on multiple directories receiving multiple events"() {
         given:
-        def dir1 = tmpDir.newFolder()
+        def latency = 0.3
         def dir2 = tmpDir.newFolder()
-        def changes = startWatcher(dir1.absolutePath, dir2.absolutePath)
+
+        println "-> $dir"
```
gavra0: nit: Remove println.

src/test/groovy/net/rubygrapefruit/platform/file/OsxFileEventsTest.groovy
```diff
@@ -23,149 +23,164 @@ import org.junit.Rule
...
-        new File(dir, "b.txt").createNewFile()
-        waitForFileSystem()
-
-        then: "change is not reported"
-        changes == [dir.canonicalPath + "/"]
-
-        when: "directory is changed after latency period"
+        when:
+        expectedChanges = expectThat pathChanged(dir)
         Thread.sleep((long) (latency * 1000 + 100))
```
gavra0: Does it need this considering .await() is invoked on line 88?

src/test/groovy/net/rubygrapefruit/platform/file/OsxFileEventsTest.groovy
```diff
@@ -23,149 +23,164 @@ import org.junit.Rule
...
-        waitForFileSystem()
-
-        then: "change is reported"
-        changes == [dir.canonicalPath + "/", dir.canonicalPath + "/"]
+        new File(dir, "b.txt").createNewFile()
 
-        cleanup:
-        stopWatcher()
+        then:
+        expectedChanges.await()
```
gavra0: Is there a way to set timeout for await()?

src/test/groovy/net/rubygrapefruit/platform/file/OsxFileEventsTest.groovy
lptr: Yes, because macOS won't report another change until `latency` is elapsed.

src/test/groovy/net/rubygrapefruit/platform/file/OsxFileEventsTest.groovy
lptr: You can pass it an explicit time span, by default it uses 1 second. I found that to be good here.

src/test/groovy/net/rubygrapefruit/platform/file/OsxFileEventsTest.groovy
lptr: Removed.

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -59,93 +61,113 @@ static void callback(ConstFSEventStreamRef streamRef,
...
-    if (pthread_create(&watcherThread, NULL, EventProcessingThread, watcherStream) != 0) {
+    watch_details_t* details = (watch_details_t*)malloc(sizeof(watch_details_t));
+    details->rootsToWatch = rootsToWatch;
+    details->watcherStream = watcherStream;
+    details->watcherCallback = watcherCallback;
+
+    if (pthread_create(&(details->watcherThread), NULL, EventProcessingThread, details) != 0) {
         mark_failed_with_errno(env, "Could not create file watcher thread.", result);
-        return;
+        return NULL;
```
adammurdoch: should probably free `details` here

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -59,93 +61,113 @@ static void callback(ConstFSEventStreamRef streamRef,
...
     }
 
+    // TODO Should this be somewhere global?
     int jvmStatus = env->GetJavaVM(&jvm);
     if (jvmStatus < 0) {
         mark_failed_with_errno(env, "Could not store jvm instance.", result);
-        return;
+        return NULL;
```
adammurdoch: should probably kill the watcher thread and free `details` here

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -166,6 +189,7 @@ Java_net_rubygrapefruit_platform_internal_jni_OsxFileEventFunctions_stopWatch(JN
     // TODO: consider using FSEventStreamFlushSync to flush all pending events.
     watcherStream = NULL;
 
+    // TODO Shouldn't we stop this first and then release the rest?
```
adammurdoch: yeah, I would stop all the work first and then free everything after

src/main/java/net/rubygrapefruit/platform/file/FileWatcherCallback.java
```diff
@@ -0,0 +1,10 @@
+package net.rubygrapefruit.platform.file;
+
+/**
+ * A callback that is invoked whenever a path has changed.
+ */
+public interface FileWatcherCallback {
+    // Invoked from native code
+    @SuppressWarnings("unused")
+    void pathChanged(String path);
```
adammurdoch: It would be good to clarify the contract here a bit more:

- Are these absolute paths?
- What happens when something is renamed or moved? deleted? created?
- Does the callback implementation need to care about thread safety?

Also, what happens when the callback throws an exception?

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
```diff
@@ -16,10 +16,55 @@
...
 
-public class OsxFileEventFunctions {
-    public static native void createWatch(String path, FunctionResult result);
-    public static native void startWatch(DefaultOsxFileEventFunctions.ChangeCallback callback, FunctionResult result);
-    public static native void stopWatch(FunctionResult result);
+import java.util.Collection;
+
+public class OsxFileEventFunctions implements NativeIntegration {
+
+    public FileWatcher startWatching(Collection<String> paths, double latency, FileWatcherCallback callback) {
```
adammurdoch: I would use an int or long and a `TimeUnit` to express the latency instead.

src/main/java/net/rubygrapefruit/platform/file/FileWatcher.java
```diff
@@ -0,0 +1,20 @@
...
+    FileWatcher EMPTY = new FileWatcher() {
+        @Override
+        public void close() {}
+    };
+
+    /**
+     * Closes the watcher and releases any native resources.
+     */
+    @Override
+    void close() throws IOException;
```
adammurdoch: Should clarify what happens when `close()` is called and there are some pending notifications or the callback is currently being executed.

src/main/cpp/apple_fsnotifier.cpp
lptr: Fixed.

src/main/cpp/apple_fsnotifier.cpp
lptr: Fixed.

src/main/cpp/apple_fsnotifier.cpp
lptr: Fixed.

src/main/java/net/rubygrapefruit/platform/file/FileWatcherCallback.java
lptr: Added docs for all this. Callback exceptions are currently ignored, need to fix that eventually.

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
lptr: Done.

src/main/java/net/rubygrapefruit/platform/file/FileWatcher.java
lptr: Clarified.

lptr:
@adammurdoch thanks for the review, I've addressed your comments in #53.

PR #52, opened 2019-12-11
More improvements to native watchers

lptr:


src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -69,15 +75,29 @@ static void callback(ConstFSEventStreamRef streamRef,
...
+        printf("~~~~ Event flags: 0x%x for %s\n", flags, paths[i]);
+        const char *type;
+        if (IS_SET(flags, kFSEventStreamEventFlagMustScanSubDirs)) {
+            type = "INVALIDATE";
+        } else if (IS_SET(flags, kFSEventStreamEventFlagItemRenamed)) {
+            if (IS_SET(flags, kFSEventStreamEventFlagItemCreated)) {
+                type = "REMOVED";
+            } else {
+                type = "CREATED";
+            }
```
lptr: This is really weird. For renames it seems that there are two events:

1) `kFSEventStreamEventFlagItemRenamed` + `kFSEventStreamEventFlagItemCreated` for the old path (i.e. the one that is not available anymore)
2) `kFSEventStreamEventFlagItemRenamed` only for the new path (i.e. that is now available).

For me the reverse would be more logical, and maybe using `kFSEventStreamEventFlagItemRemoved` for the old path would be even more logical. Mysterious!

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -25,7 +25,13 @@ typedef struct watch_details {
     CFRunLoopRef threadLoop;
 } watch_details_t;
 
-static void reportEvent(const char *event, char *path, jobject watcherCallback) {
+static jobject getTypeEnum(JNIEnv *env, const char *name) {
```
adammurdoch: generally, what we do elsewhere is do this lookup in Java land and instead just pass the ordinals for the constants across the JNI boundary.

lptr: Changed it so in #53. Thanks for the review.

PR #53, opened 2019-12-12
File system watcher improvements

lptr:
Addressing some review comments from #50, #51 and #52.


src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -135,57 +165,61 @@ Java_net_rubygrapefruit_platform_internal_jni_OsxFileEventFunctions_startWatchin
         mark_failed_with_errno(env, "No paths given to watch.", result);
         return NULL;
     }
+
+    watch_details_t* details = (watch_details_t*)malloc(sizeof(watch_details_t));
```
```suggestion
    watch_details_t* details = (watch_details_t*) malloc(sizeof(watch_details_t));
```

Nit

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
```diff
@@ -0,0 +1,96 @@
...
+        return watcher;
+    }
+
+    /**
+     * Canonicalizes the given paths using {@link File#getCanonicalPath()}.
+     * Throws {@link NativeException} if any of the given paths is not absolute,
+     * or if they cannot be canonicalized for any reason.
+     */
+    private static List<String> canonicalizeAbsolutePaths(Collection<String> watchRoots) {
+        List<String> canonicalPaths = new ArrayList<String>(watchRoots.size());
```
wolfs: It seems to me that this should be done by the client of the method. WDYT?

src/main/java/net/rubygrapefruit/platform/internal/jni/WindowsFileEventFunctions.java
```diff
@@ -16,53 +16,67 @@
...
+     *
+     *     <li>Events arrive from a single background thread unique to the {@link FileWatcher}.</li>
+     *
+     *     <li>Removals are reported as a
+     *     {@link net.rubygrapefruit.platform.file.FileWatcherCallback.Type#MODIFIED MODIFIED} and a
+     *     {@link net.rubygrapefruit.platform.file.FileWatcherCallback.Type#REMOVED REMOVED} event.</li>
+     *
+     *     <li>Renames are reported as the source file being
+     *     {@link net.rubygrapefruit.platform.file.FileWatcherCallback.Type#REMOVED REMOVED}.
+     *     The creation of the target file is not reported.</li>
```
wolfs: Ouch. How can we handle that on the Gradle side?

lptr: Re-read the whole directory every time? :) This is what we do now in any case, so it works.

We can either spend some time to figure out how to make Windows generate an event for the created file, too, or we can post some special event for renames, and... _re-read the whole directory_ on the Gradle side.

I made a note to do these things before we go into production.

src/main/cpp/apple_fsnotifier.cpp
lptr: Fixed.

src/main/java/net/rubygrapefruit/platform/internal/jni/WindowsFileEventFunctions.java
wolfs: Which directory? I thought this was about a move to a different directory. Do when only get the `REMOVED` event and not the `CREATED` event in the new location?

lptr: This is about renaming inside the same watched directory. It’s not move, it’s literally rename. Moving stuff out of the watched dir or into it works well.

adammurdoch: Perhaps we should clarify these behaviours in the comment, and document that "rename" means "change of file name within the same directory" (which wasn't immediately obvious to me) and perhaps also what happens for things like "move within the watched directory" and "moved into watched directory" and "moved out of watched directory"

PR #54, opened 2019-12-12
Handle locking file event on macOS

wolfs:
When Gradle locks a file, the macOS watchers receive a `0x10400`event. That is `kFSEventStreamEventFlagItemInodeMetaMod | kFSEventStreamEventFlagItemIsFile `.

This currently causes Gradle to drop the whole VFS state when used on the `gradle/gradle` build.

This PR hands on the change as `MODIFIED` to the Java side. Another option would be to completely ignore the event.

### DIFF
--- a/src/main/cpp/apple_fsnotifier.cpp
+++ b/src/main/cpp/apple_fsnotifier.cpp
@@ -86,6 +86,9 @@ static void callback(ConstFSEventStreamRef streamRef,
             type = FILE_EVENT_REMOVED;
         } else if (IS_SET(flags, kFSEventStreamEventFlagItemCreated)) {
             type = FILE_EVENT_CREATED;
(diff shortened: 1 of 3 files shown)


wolfs:
I wonder if we should ignore unknown events in general. Or at least handle all the constants from the documentation in some way.

wolfs:
I'll merge this now. Let's polish it on a separate PR if necessary.

adammurdoch:
I think it would be better if we changed the APIs so that we either always ignore events that don't affect the _content_ of the watched tree, or we fire the event but indicate that the content did not change.

wolfs:
+1 We should try to handle all events we know about and write tests for them. E.g. there also seems to be an event type `kFSEventStreamEventFlagItemChangeOwner` which indicates that maybe the owner of a file changed?

lptr:
Hm. This did happen to me once when I tried it and wanted to look up what the problem was. But for the rest of the builds I tried it never occurred again.

PR #55, opened 2019-12-13
Add --watch to test app

wolfs:
So we can easily try out watching a directory on macOs or Windows.

### DIFF
--- a/test-app/src/main/java/net/rubygrapefruit/platform/test/Main.java
+++ b/test-app/src/main/java/net/rubygrapefruit/platform/test/Main.java
@@ -19,21 +19,40 @@ package net.rubygrapefruit.platform.test;
 import joptsimple.OptionException;
 import joptsimple.OptionParser;
 import joptsimple.OptionSet;
-import net.rubygrapefruit.platform.*;
+import net.rubygrapefruit.platform.Native;
+import net.rubygrapefruit.platform.NativeIntegrationUnavailableException;
 import net.rubygrapefruit.platform.Process;
-import net.rubygrapefruit.platform.file.*;
+import net.rubygrapefruit.platform.SystemInfo;
+import net.rubygrapefruit.platform.file.DirEntry;
+import net.rubygrapefruit.platform.file.FileInfo;
+import net.rubygrapefruit.platform.file.FileSystemInfo;
+import net.rubygrapefruit.platform.file.FileSystems;
+import net.rubygrapefruit.platform.file.FileWatcher;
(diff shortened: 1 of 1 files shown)


test-app/src/main/java/net/rubygrapefruit/platform/test/Main.java
```diff
@@ -347,6 +372,49 @@ private static void fileSystems() {
...
+        }
+        try {
+            System.out.println("Waiting - type ctrl-d to exit ...");
+            while (true) {
+                int ch = System.in.read();
+                if (ch < 0) {
+                    break;
+                }
+            }
+            watcher.close();
```
lptr: Why close twice?

wolfs: Closing only once now.

lptr:
Now that https://github.com/adammurdoch/native-platform/pull/58 is merged we can remove the `System.exit()`.

wolfs:
I removed the `System.exit()` call.

test-app/src/main/java/net/rubygrapefruit/platform/test/Main.java
```diff
@@ -19,21 +19,40 @@
...
+import net.rubygrapefruit.platform.file.FileSystemInfo;
+import net.rubygrapefruit.platform.file.FileSystems;
+import net.rubygrapefruit.platform.file.FileWatcher;
+import net.rubygrapefruit.platform.file.FileWatcherCallback;
+import net.rubygrapefruit.platform.file.Files;
+import net.rubygrapefruit.platform.file.PosixFileInfo;
+import net.rubygrapefruit.platform.file.PosixFiles;
+import net.rubygrapefruit.platform.internal.Platform;
+import net.rubygrapefruit.platform.internal.jni.OsxFileEventFunctions;
+import net.rubygrapefruit.platform.internal.jni.WindowsFileEventFunctions;
```
adammurdoch: I don't think we should add this until there is a public API for file events, as this is also a sample and documentation for the APIs.

wolfs: So you think we should remove it from `master` again for the time being? It was very helpful to use the test-app to debug problems with file watching.

adammurdoch: That's one option. The other is to add a public API, which is what I would do. We just need to extract an interface or two out of the implementations.

PR #57, opened 2019-12-13
Detach native thread from JVM at the end

wolfs:
Instead of trying to attach the thread to the JVM every time we report a change, we only attach it when the thread starts and detach it at the end so the JVM can exit.


lptr:
Handled in https://github.com/adammurdoch/native-platform/pull/58.

PR #58, opened 2019-12-13
Better JNI thread management

lptr:
Instead of trying to attach the thread to the JVM every time we report a change, we only attach it when the thread starts and detach it at the end so the JVM can exit.

Fixes https://github.com/adammurdoch/native-platform/issues/56.

### DIFF
--- a/src/main/cpp/apple_fsnotifier.cpp
+++ b/src/main/cpp/apple_fsnotifier.cpp
@@ -14,13 +14,13 @@
 #include <pthread.h>
 #include <strings.h>
 
-JavaVM* jvm = NULL;
 bool invalidStateDetected = false;
 
 typedef struct watch_details {
     CFMutableArrayRef rootsToWatch;
     FSEventStreamRef watcherStream;
     pthread_t watcherThread;
+    JavaVM* jvm;
     JNIEnv *env;
     jobject watcherCallback;
     CFRunLoopRef threadLoop;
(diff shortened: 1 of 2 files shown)


src/main/cpp/win_fsnotifier.cpp
```diff
@@ -148,9 +151,9 @@ Java_net_rubygrapefruit_platform_internal_jni_WindowsFileEventFunctions_startWat
 
     printf("\n~~~~ Configuring...\n");
 
-    // TODO Should this be somewhere global?
+    JavaVM* jvm;
     int jvmStatus = env->GetJavaVM(&jvm);
```
lptr: Interestingly, `GetJavaVM()` is supposed to return a `JavaVM` _for the current thread,_ yet we can happily reuse it in the watcher thread, too. There also doesn't seem to be another way to get hold of a `JavaVM` instance, save for starting a new JVM. Maybe the documentation is incorrect?

> Returns the Java VM interface (used in the Invocation API) associated with the current thread. The result is placed at the location pointed to by the second argument, vm.

https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/functions.html#wp23168

lptr: Ah, it doesn't actually say that the `JavaVM` needs to be associated with only one thread. OK, I guess we use it the right way then.

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -14,18 +14,19 @@
...
 #include <strings.h>
 
-JavaVM* jvm = NULL;
 bool invalidStateDetected = false;
 
 typedef struct watch_details {
     CFMutableArrayRef rootsToWatch;
     FSEventStreamRef watcherStream;
     pthread_t watcherThread;
+    JavaVM* jvm;
```
```suggestion
    JavaVM *jvm;
```

Nit.

PR #61, opened 2020-01-16
macOS handle watched directory hierarchy deleted event

lptr:
We didn't handle the event when a watched directory was deleted in its entirety. Now we do.


lptr:
And now need to fix the tests for Windows.

PR #63, opened 2020-01-23
Use calculated native version

wolfs:
Instead of using a manual version for the native Library,
the version is generated from the hashes of the native sources.
This way, the version is different whenever the native libraries change and the cache-directory to where the unpacking happens is safe to use
even for snapshot versions.


src/main/java/net/rubygrapefruit/platform/Native.java
```diff
@@ -1,98 +1,99 @@
...
+
+import java.io.File;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Provides access to the native integrations. Use {@link #get(Class)} to load a particular integration.
+ */
+@ThreadSafe
+public class Native {
```
wolfs: Looks like this file got converted from CRLF to LF line endings...

lptr: Can you split that to a separate commit so the actual changes show up?

build.gradle
```diff
@@ -67,6 +67,22 @@ task nativeHeaders {
...
 
+def writeNativeVersionSources = tasks.register("writeNativeVersionSources", WriteNativeVersionSources) {
+    ["main", "shared", "curses"].each {
+        nativeSources.from("src/${it}/cpp")
+        nativeSources.from("src/${it}/headers")
+    }
+
+    def generatedSources = project.layout.buildDirectory.dir("generated/sources")
+
+    nativeVersionHeaderFile = file("$nativeHeadersDir/native_platform_version.h")
```
lptr: We should not put this into the same directory as the output of `nativeHeaders`: that's overlapping outputs!

build.gradle
```diff
@@ -67,6 +67,22 @@ task nativeHeaders {
     }
 }
 
+def writeNativeVersionSources = tasks.register("writeNativeVersionSources", WriteNativeVersionSources) {
+    ["main", "shared", "curses"].each {
+        nativeSources.from("src/${it}/cpp")
+        nativeSources.from("src/${it}/headers")
+    }
```
lptr: I wish we had source sets for C++ as well so we could reference them here instead of redeclaring the paths twice.

build.gradle
```diff
@@ -350,8 +366,12 @@ model {
...
 }
 
-tasks.withType(CppCompile) { task ->
-    task.dependsOn project.tasks.nativeHeaders
+tasks.withType(CppCompile).configureEach {
+    dependsOn(project.tasks.nativeHeaders)
+    dependsOn(writeNativeVersionSources)
+    inputs.files(nativeHeadersDir)
+        .withPropertyName("nativeHeadersDir")
+        .withPathSensitivity(PathSensitivity.RELATIVE)
```
lptr: Why do we manually have to add this as a dependency/input files using the runtime API? Shouldn't these be specified as input directories? And shouldn't we use the output of the tasks instead of declaring the dependencies separately?

build.gradle
wolfs: I think there are, I just don't know where...

build.gradle
wolfs: This is not overlapping outputs, since both tasks writing to the directory have an output file. The parent directory is not an output of any of the tasks.

build.gradle
wolfs: The directory doesn't know that the two files in the directory are generated by two separate tasks.

src/main/java/net/rubygrapefruit/platform/Native.java
wolfs: I managed to fix it. My global `core.autocrlf` git config was `input`. I set it to false and did a new checkout of native platform.

I think it would make sense to convert all files (or at least all Java files) to Unix line endings, WDYT?

build.gradle
wolfs: So the inputs are now detected automatically by header analysis. It seems like the header analysis loses the task dependencies, since `CalculatedTaskInputFileCollection` doesn't override `visitDependencies`.

wolfs:
@lptr PTAL!

build.gradle
lptr: Ah, that's true. I think I'd still feel better if we had it in a separate directory.

wolfs:
We can't use task outputs as long as https://github.com/gradle/gradle/issues/11998 isn't fixed and we don't want to add the inputs twice.

build.gradle
wolfs: Both tasks now output to a separate directory.

PR #65, opened 2020-01-24
Allow building for 32bits on Linux always

wolfs:


build.gradle
```diff
@@ -300,8 +300,7 @@ model {
 
             binaries.withType(SharedLibraryBinarySpec) { binary ->
                 def targetOs = binary.targetPlatform.operatingSystem
-                def targetArch = binary.targetPlatform.architecture
-                def includeVariant = !onlyPrimaryVariants || targetOs.windows || targetOs.macOsX || (targetOs.linux && targetArch.name == "x86-64" && !targetPlatform.name.contains("ncurses6"))
+                def includeVariant = !onlyPrimaryVariants || targetOs.windows || targetOs.macOsX || (targetOs.linux && (targetPlatform.architecture.i386 || targetPlatform.architecture.amd64) && !targetPlatform.name.contains("ncurses6"))
```
lptr: Let's split these very long lines for better readability and to show structure.

build.gradle
```diff
@@ -269,7 +269,7 @@ model {
 
         all {
             binaries.all {
-                if (targetPlatform.operatingSystem.name in ['linux', 'freebsd'] && targetPlatform.architecture != arch) {
+                if (targetPlatform.operatingSystem.name in ['linux', 'freebsd'] && targetPlatform.architecture != arch && !(arch.amd64 && targetPlatform.architecture.i386)) {
```
adammurdoch: I don't think we should just assume that we can build the x86 variant, because often the required dependencies are missing. Some options:

- probe for the library
- fix the native plugins to probe for the library
- add some opt-in flag
- assume, but probe for the library and fail if not present

wolfs:
Closing, since we are not supporting 32bit any more: https://github.com/gradle/native-platform/pull/100

PR #68, opened 2020-01-27
Relax Java requirement for building native platform

wolfs:
In order to use native platform in a composite build with `gradle/gradle`, the Java version for building needs to be relaxed, since `gradle/gradle` requires Jav 11 to build.

For Java 11, the `javah` binary is not shipped anymore, though `javac` can be used directly to generate the JNI headers, even with Java 8. Therefore, this PR changes the JNI header generation to use `javac -h` and then concatenates the generated headers into one `native.h` as before.


buildSrc/src/main/java/ConcatenateJniHeaders.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public abstract class ConcatenateJniHeaders extends DefaultTask {
+
+    /**
+     * The directory containing the generated JNI headers.
+     *
+     * Contains one header file per class with a native method.
+     */
+    @InputDirectory
+    public abstract DirectoryProperty getJniHeaders();
```
adammurdoch: This should really be a file collection, as the task does not really care where the files are or how they are arranged. It just needs a set of files.

buildSrc/src/main/java/ConcatenateJniHeaders.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * The target directory for the concatenated header file.
+     *
+     * The file will be called <pre>native.h</pre>
+     */
+    @OutputDirectory
+    public abstract DirectoryProperty getGeneratedNativeHeaderDirectory();
```
adammurdoch: This should really be a file, as the task produces a single file, not a directory hierarchy.

lptr: 👍 I think so too. We should still configure each header-generating task to produce output into its own directory.

wolfs: Is there a way to get the parent dir of the output file while retaining task dependency information? This directory would be used as the include directory for native compilation.

buildSrc/src/main/java/ConcatenateJniHeaders.java
wolfs: It is a `ConfigurableFileCollection` now.

PR #69, opened 2020-01-28
Convert one variant to use the new plugins

wolfs:


adammurdoch:
I'm not sure I'd spend too much time trying to go down this path at the moment. The old plugins are working fine and will do what we need to do for the foreseeable future. Of course, it would be better to use the new plugins if they supported what we need, but I don't think that's the most important thing for now.

wolfs:
We're are not going to do this right now and composite build works now via a workaround.

PR #70, opened 2020-01-28
Allow depending only on the variants which can be built locally

wolfs:
So a working version can be published to the local Maven repository
and used from another build.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,7 +34,10 @@ dependencies {
     testImplementation 'org.spockframework:spock-core:1.0-groovy-2.4'
 }
 
+// Only depend on variants which can be built on our CI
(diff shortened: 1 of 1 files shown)


wolfs:
Doing the following works for me:
- run `./gradlew publishToMavenLocal -PonlyLocalVariants` from this branch
- add the `mavenLocal()` repository to the `gradle/gradle` build and change the version of native platform in `dependencies.gradle` to `0.22-dev`.

As a result, I can run integration tests with the native version just built on my machine: https://e.grdev.net/s/usp3vgujbydh4/dependencies?dependencies=native-platform&expandAll

lptr:
Sounds great. Let's make sure this gets documented on the main `README.md`.

PR #71, opened 2020-01-28
Allow native code to log via JUL

lptr:
Log using `java.util.logging`.


src/main/java/net/rubygrapefruit/platform/Native.java
```diff
@@ -1,99 +1,101 @@
...
- *    Unless required by applicable law or agreed to in writing, software
- *    distributed under the License is distributed on an "AS IS" BASIS,
- *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- *    See the License for the specific language governing permissions and
- *    limitations under the License.
- */
-
-package net.rubygrapefruit.platform;
-
-import net.rubygrapefruit.platform.internal.NativeLibraryLoader;
```
wolfs: I suppose this file now also has different line endings. Shall we convert all files to LF on `master` in a separate PR?

lptr: Damn. Yeah, let's convert. Would you mind taking that on?

wolfs: Done.

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -198,7 +198,7 @@ Java_net_rubygrapefruit_platform_internal_jni_OsxFileEventFunctions_startWatchin
     for (int i = 0; i < count; i++) {
         jstring path = (jstring) env->GetObjectArrayElement(paths, i);
         char* watchedPath = java_to_char(env, path, result);
-        printf("~~~~ Watching %s\n", watchedPath);
+        log_info(env, "~~~~ Watching %s\n", watchedPath);
```
wolfs: I suppose we don't need the `~~~~` any more, right?

src/main/java/net/rubygrapefruit/platform/Native.java
```diff
@@ -1,99 +1,101 @@
...
+            if (loader == null) {
+                Platform platform = Platform.current();
+                try {
+                    loader = new NativeLibraryLoader(platform, new NativeLibraryLocator(extractDir));
+                    loader.load(platform.getLibraryName(), platform.getLibraryVariants());
+                    long nativeVersion = NativeLibraryFunctions.getVersion();
+                    if (nativeVersion != NativeVersion.VERSION) {
+                        throw new NativeException(String.format("Unexpected native library version loaded. Expected %s, was %s.", NativeVersion.VERSION, nativeVersion));
+                    }
+                    NativeLogger.initLogging(Native.class);
```
wolfs: I wonder if setting the log level at this point is a good idea. For Gradle, is is a Global service, so the logging probably isn't initialized, yet, right? Especially it won't be initialized between builds, so if I run one build with `--debug` and one with `--quiet`, the log level of the native part won't change I suppose. Or am I understanding something wrong here?

wolfs:
Some test coverage would be nice as well.

src/main/cpp/apple_fsnotifier.cpp
lptr: I'll rewrite these soon, and we can change the log messages then.

src/main/java/net/rubygrapefruit/platform/Native.java
lptr: Let's fix this separately. Seems like the change on the Gradle side is relatively big.

PR #74, opened 2020-01-29
Add documentation how to test integration with Gradle

wolfs:


lptr:
Can we put the Gradle-side stuff in permanently, so they wouldn't need mentioning here?

wolfs:
> Can we put the Gradle-side stuff in permanently, so they wouldn't need mentioning here?

What do you mean by the "Gradle-side stuff"? The only which could be done on the Gradle side is to add `mavenLocal()`, and I don't think that is a good idea. Changing the native platform version is necessary dependent on what snapshot/dev build you are using.

wolfs:
@lptr PTAL!

lptr:
NB: it's `--include-build` not `--includeBuild`.

wolfs:
@lptr PTAL!

PR #75, opened 2020-01-29
Publish commit status to Github

wolfs:
and cleanup TC DSL.

### DIFF
--- a/.teamcity/NativePlatformBuild.kt
+++ b/.teamcity/NativePlatformBuild.kt
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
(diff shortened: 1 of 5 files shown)


wolfs:
@lptr Please take a look.

lptr:
Looks like a lot of code. Is there a way to publish a reusable plugin with the parts that are the same across our builds?

wolfs:
@lptr The changes for publishing the status to github is only this: https://github.com/gradle/native-platform/pull/75/commits/44783b5f927e455780d86c08c50e48e99be1daa8

The remaining changes are about going from the auto-generated Kotlin DSL to something better manageable. That part is not really reusable, given that it is specific to how native platform is being built.

PR #76, opened 2020-01-29
Add native publication in current model

wolfs:
This allows composite build to work when used together
with native platform

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -331,7 +331,10 @@ publishing {
     }
 }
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -369,6 +331,49 @@ publishing {
     }
 }
 
+afterEvaluate {
```
lptr: Let's add some comments about a) why we are doing this here now, and b) how to revert if needed/in case we move to the non-software model plugins.

wolfs: I added some comments.

PR #77, opened 2020-01-30
Directly include generated headers

wolfs:
instead of concatenating them into one file.


buildSrc/src/main/java/JniPlugin.java
```diff
@@ -23,34 +21,19 @@
...
                 tasks,
-                compileJavaProvider,
-                project.files(compilerArguments.getGeneratedHeadersDirectory()).getAsFileTree(),
-                jniExtension.getGeneratedHeader()
-            );
-
-            configureIncludePath(tasks, concatenateJniHeaders
-                .map(ConcatenateJniHeaders::getGeneratedNativeHeader)
-                .map(it -> it.get().getAsFile().getParentFile())
+                compileJavaProvider.map(compileJava -> compilerArguments.getGeneratedHeadersDirectory())
```
lptr: I don't understand this mapping here. `compileJava` is not used at all, can we use `compilerArguments.getGeneratedHeadersDirectory()` directly maybe?

src/curses/cpp/curses.cpp
```diff
@@ -19,7 +19,7 @@
  */
 #ifndef _WIN32
 
-#include "native.h"
+#include "net_rubygrapefruit_platform_internal_jni_TerminfoFunctions.h"
```
lptr: OK, that's ugly :o

buildSrc/src/main/java/JniPlugin.java
wolfs: Yeah, I use this "trick" to add the task dependency to the resulting provider. `compilerArguments.getGeneratedHeadersDirectory()` currently doesn't carry the dependency to the task since it is in a `@Nested` property.

wolfs:
It seems like compilation doesn't fail even when a JNI header is not included, not even for the current platform. It only fails at runtime, when the test tries to invoke the native method.

buildSrc/src/main/java/JniPlugin.java
wolfs: I simplified the code, given that the task dependency doesn't work anyway for include directories.

wolfs:
@lptr PTAL!

PR #78, opened 2020-01-31
Use automatic task dependencies

wolfs:
It seems like the manual `dependsOn` calls are mostly unnecessary.


buildSrc/src/main/java/JniPlugin.java
```diff
@@ -1,92 +1,38 @@
...
-        RemoveGeneratedNativeHeaders removeGeneratedNativeHeaders,
         TaskProvider<JavaCompile> compileJavaProvider
     ) {
         compileJavaProvider.configure(compileJava -> {
-            compileJava.getOptions().getCompilerArgumentProviders().add(compilerArguments);
-            // Cannot do incremental header generation
+            DirectoryProperty headerOutputDirectory = compileJava.getOptions().getHeaderOutputDirectory();
+            headerOutputDirectory.convention(compileJava.getProject().getLayout().getBuildDirectory().dir("generated/jni-headers"));
+            // The nested output is not marked automatically as an output of the task regarding task dependencies.
+            // So we mark it manually here.
```
lptr: Can we link to an issue here?

buildSrc/src/main/java/JniPlugin.java
```diff
@@ -1,92 +1,38 @@
...
+            compileJava.getOutputs().dir(compileJava.getOptions().getHeaderOutputDirectory());
+            // Cannot do incremental header generation, since the pattern is currently wrong.
```
lptr: Link to issue here too?

buildSrc/src/main/java/JniPlugin.java
wolfs: Added.

buildSrc/src/main/java/JniPlugin.java
wolfs: Added.

wolfs:
@lptr PTAL!

lptr:
LGTM.

PR #79, opened 2020-02-04
Publish FreeBSD variant from CI

wolfs:
Now that there are FreeBSD agents, we also can publish the variant from CI.


wolfs:
Seems to work: https://repo.gradle.org/gradle/webapp/#/artifacts/browse/tree/General/ext-snapshots-local/net/rubygrapefruit/native-platform-freebsd-amd64-libcpp/0.22-snapshot-20200204104842+0000

PR #80, opened 2020-02-04
Show published version in Teamcity build status

wolfs:


wolfs:
This is what it looks like:
image

PR #83, opened 2020-02-06
Upgrade to Spock 1.3

lptr:
This is to fix compiling tests using Java 9+.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,8 +41,7 @@ allprojects {
 }
 
 dependencies {
-    testImplementation 'org.codehaus.groovy:groovy-all:2.4.1'
-    testImplementation 'org.spockframework:spock-core:1.0-groovy-2.4'
+    testImplementation 'org.spockframework:spock-core:1.3-groovy-2.5'
 }
 
 // Only depend on variants which can be built on our CI
--- a/src/test/groovy/net/rubygrapefruit/platform/testfixture/JavaVersion.groovy
+++ b/src/test/groovy/net/rubygrapefruit/platform/testfixture/JavaVersion.groovy
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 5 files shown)


src/test/groovy/net/rubygrapefruit/platform/testfixture/JavaVersion.groovy
```diff
@@ -0,0 +1,82 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.rubygrapefruit.platform.testfixture
+
+class JavaVersion {
```
wolfs: Do we need all this complexity? Is there an easier way to get the major version of Java?

build.gradle
```diff
@@ -41,8 +41,7 @@ allprojects {
 }
 
 dependencies {
-    testImplementation 'org.codehaus.groovy:groovy-all:2.4.1'
-    testImplementation 'org.spockframework:spock-core:1.0-groovy-2.4'
+    testImplementation 'org.spockframework:spock-core:1.3-groovy-2.5'
```
wolfs: Spock 1.3 requires Java 8. That means that the tests can't run on anything < Java 8. Since we seem to only manually test for anything less than Java 8 that could be fine.

wolfs: Looks like I misread the sentence. Spock 1.2 and 1.3 require Java 7 (they drop support for Java 6): http://spockframework.org/spock/docs/1.3/release_notes.html#_1_2_2018_09_23

Still something to consider.

lptr: @adammurdoch can we drop automated testing for Java 5 and 6?

I'm going to merge this for now, and if Adam disagrees we can revert. It's not a big deal either way.

src/test/groovy/net/rubygrapefruit/platform/testfixture/JavaVersion.groovy
lptr: Probably, but this code is copied from Gradle, and works well. It's only test code, too.

BTW, I think "silently do nothing on Java 11+" is not a good production behavior for the feature tested here. The code should emit some warning or fail; something we can actually test.

We could also deprecate this functionality in `native-platform` and remove in a future version.

PR #85, opened 2020-02-10
Listen to all events on macOS

lptr:
We've missed a few up until now.

### DIFF
--- a/src/main/cpp/apple_fsnotifier.cpp
+++ b/src/main/cpp/apple_fsnotifier.cpp
@@ -163,7 +163,7 @@ void freeDetails(JNIEnv *env, watch_details_t *details) {
 JNIEXPORT jobject JNICALL
 Java_net_rubygrapefruit_platform_internal_jni_OsxFileEventFunctions_startWatching(JNIEnv *env, jclass target, jobjectArray paths, long latencyInMillis, jobject javaCallback, jobject result) {
 
-    log_fine(env, "\nConfiguring...", NULL);
+    log_fine(env, "Configuring...", NULL);
 
     invalidStateDetected = false;
 
--- a/src/main/cpp/apple_fsnotifier.cpp
+++ b/src/main/cpp/apple_fsnotifier.cpp
@@ -40,7 +40,7 @@ static void reportEvent(jint type, char *path, watch_details_t *details) {
         }
     }
 
-    log_info(details->env, "~~~~ Changed: %s %d", path, type);
+    log_info(details->env, "Changed: %s %d", path, type);
 
     JNIEnv *env = details->env;
     jobject watcherCallback = details->watcherCallback;
@@ -62,7 +62,7 @@ static void callback(ConstFSEventStreamRef streamRef,
 
     for (int i = 0; i < numEvents; i++) {
         FSEventStreamEventFlags flags = eventFlags[i];
-        log_fine(details->env, "~~~~ Event flags: 0x%x for %s", flags, paths[i]);
+        log_fine(details->env, "Event flags: 0x%x for %s", flags, paths[i]);
         jint type;
         if (IS_SET(flags, kFSEventStreamEventFlagHistoryDone)) {
             continue;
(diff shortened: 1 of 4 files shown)


src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -78,13 +84,14 @@ static void callback(ConstFSEventStreamRef streamRef,
...
         } else if (IS_SET(flags, kFSEventStreamEventFlagItemCreated)) {
             type = FILE_EVENT_CREATED;
-        } else if (IS_SET(flags, kFSEventStreamEventFlagItemInodeMetaMod)) {
-            // File locked
+        } else if (IS_ANY_SET(flags,
+                kFSEventStreamEventFlagItemInodeMetaMod // file locked
+                | kFSEventStreamEventFlagItemFinderInfoMod
+                | kFSEventStreamEventFlagItemChangeOwner
+                | kFSEventStreamEventFlagItemXattrMod)) {
             type = FILE_EVENT_MODIFIED;
```
wolfs: Shall we say `METADATA_MODIFIED` instead?

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -62,9 +62,15 @@ static void callback(ConstFSEventStreamRef streamRef,
...
     for (int i = 0; i < numEvents; i++) {
         FSEventStreamEventFlags flags = eventFlags[i];
-        log_fine(details->env, "~~~~ Event flags: 0x%x for %s\n", flags, paths[i]);
+        log_fine(details->env, "Event flags: 0x%x for %s", flags, paths[i]);
         jint type;
-        if (IS_SET(flags, kFSEventStreamEventFlagMustScanSubDirs)) {
+        if (IS_SET(flags, kFSEventStreamEventFlagHistoryDone)) {
+            continue;
+        } else if (IS_ANY_SET(flags,
+                kFSEventStreamEventFlagRootChanged
```
wolfs: Shall we also say if the flags are equal to `kFSEventStreamEventFlagNone` - which is `0x00000000`, then we return `INVALIDATE`? IIUC, that was the legacy event before the API was able to send better events.

lptr: Hm, that's interesting. I wonder how useful that would be, though, as we'd effectively have two states: 1) nothing's changed, 2) everything's changed. This would mostly eliminate the utility of VFS retention, no? It is more or less the same outcome as we have now via `UNKNOWN`.

The more detailed API has been around since OS X Lion 10.7, which was released in 2011. I don't expect to encounter even earlier versions of macOS with VFS retention enabled. If users report a problem, we can always explicitly disable the functionality on those OS versions.

WDYT?

src/main/cpp/apple_fsnotifier.cpp
lptr: Yeah, I guess we could do that, though that's changing the API a bit.

I wonder how useful the distinction for us is in any case. Would you consider a locking event to be a metadata change or content change (assuming the file is now unavailable to be read by Gradle)? Same for owner change.

lptr: I mean, Gradle would still need to invalidate its VFS cache on `METADATA_CHANGED`, right? We can distinguish even so.

src/main/cpp/apple_fsnotifier.cpp
wolfs: That's fine, we can also leave it out.

src/main/cpp/apple_fsnotifier.cpp
wolfs: We can leave it as `MODIFIED` for now. I would return everything which doesn't change the contents as `METADATA_CHANGED`. Note that the `kFSEventStreamEventFlagItemModified` we capture above also may have some metadata change flag set.

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -78,14 +78,14 @@ static void callback(ConstFSEventStreamRef streamRef,
             } else {
                 type = FILE_EVENT_CREATED;
             }
+        } else if (IS_SET(flags, kFSEventStreamEventFlagItemModified)) {
+            type = FILE_EVENT_MODIFIED;
         } else if (IS_ANY_SET(flags,
                 kFSEventStreamEventFlagItemInodeMetaMod // file locked
```
wolfs: Let's do the check for metadata changes after the modified/created/removed changes. I think those flags trump the metadata changes.

PR #86, opened 2020-02-10
Implement native file watcher for Linux

lptr:
What it says on the tin.

### DIFF
--- a/src/file-events/headers/linux_fsnotifier.h
+++ b/src/file-events/headers/linux_fsnotifier.h
@@ -2,12 +2,49 @@
 
 #ifdef __linux__
 
-#include <list>
+#include <sys/inotify.h>
+#include <unordered_map>
 
 #include "generic_fsnotifier.h"
 #include "net_rubygrapefruit_platform_internal_jni_LinuxFileEventFunctions.h"
 #include "net_rubygrapefruit_platform_internal_jni_LinuxFileEventFunctions_WatcherImpl.h"
 
 using namespace std;
+
+class Server;
+
+class WatchPoint {
+public:
+    WatchPoint(const u16string& path, int fdInotify);
+    ~WatchPoint();
+
+    const int watchDescriptor;
+
+private:
+    const int fdInotify;
+};
+
+class Server : public AbstractServer {
+public:
+    Server(JNIEnv* env, jobject watcherCallback);
+    ~Server();
+
+    void startWatching(const u16string& path);
+    void stopWatching(const u16string& path);
+
+    // TODO This should be private
+    void handleEvents(
+        size_t numEvents);
+
+protected:
+    void runLoop(JNIEnv* env, function<void(exception_ptr)> notifyStarted) override;
+
+private:
+    void handleEvent(JNIEnv* env, char* path, int flags);
+
+    unordered_map<u16string, WatchPoint> watchPoints;
+    unordered_map<int, u16string> watchDescriptors;
+    const int fdInotify;
+};
 
 #endif
--- a/src/file-events/headers/linux_fsnotifier.h
+++ b/src/file-events/headers/linux_fsnotifier.h
@@ -7,7 +7,6 @@
 
 #include "generic_fsnotifier.h"
 #include "net_rubygrapefruit_platform_internal_jni_LinuxFileEventFunctions.h"
-#include "net_rubygrapefruit_platform_internal_jni_LinuxFileEventFunctions_WatcherImpl.h"
 
 using namespace std;
 
@@ -29,8 +28,8 @@ public:
     Server(JNIEnv* env, jobject watcherCallback);
     ~Server();
 
-    void startWatching(const u16string& path);
-    void stopWatching(const u16string& path);
+    void startWatching(const u16string& path) override;
+    void stopWatching(const u16string& path) override;
 
     // TODO This should be private
     void handleEvents(
--- a/src/file-events/headers/linux_fsnotifier.h
+++ b/src/file-events/headers/linux_fsnotifier.h
@@ -17,6 +17,8 @@ public:
     WatchPoint(const u16string& path, int fdInotify);
     ~WatchPoint();
 
+    void close();
+
     const int watchDescriptor;
 
 private:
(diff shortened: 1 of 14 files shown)


lptr:
I'm going to add some more error handling today, and then we can merge this.

src/test/groovy/net/rubygrapefruit/platform/file/AbstractFileEventsTest.groovy
```diff
@@ -551,7 +604,7 @@ abstract class AbstractFileEventsTest extends Specification {
     }
 
     @Unroll
-    @IgnoreIf({ Platform.current().windows })
+    @IgnoreIf({ Platform.current().windows || Platform.current().linux })
```
wolfs: Is this something we expect to fix for Linux and Windows? Can you add a comment?

src/test/groovy/net/rubygrapefruit/platform/file/LinuxFileEventFunctionsTest.groovy
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    protected void waitForChangeEventLatency() {
+        TimeUnit.MILLISECONDS.sleep(50)
+    }
+
+    @Override
+    protected void stopWatcher() {
+        // FIXME This is necessary to avoid a race condition
```
wolfs: Is this still necessary?

src/file-events/headers/linux_fsnotifier.h
```diff
@@ -0,0 +1,52 @@
...
+#include <sys/eventfd.h>
+#include <sys/inotify.h>
+#include <unordered_map>
+
+#include "generic_fsnotifier.h"
+#include "net_rubygrapefruit_platform_internal_jni_LinuxFileEventFunctions.h"
+
+using namespace std;
+
+class Server;
```
wolfs: I wonder if you only use the classes defined in this header file in the Linux watcher, shouldn't those be moved to `linux_fsnotifier.h`? Same for the other headers.

src/file-events/headers/linux_fsnotifier.h
```diff
@@ -0,0 +1,52 @@
...
+    void processCommandsOnThread() override;
+    void terminate() override;
+
+private:
+    void handleEventsInBuffer(JNIEnv* env, const char* buffer, ssize_t bytesRead);
+    void handleEvent(JNIEnv* env, const inotify_event* event);
+
+    unordered_map<u16string, WatchPoint> watchPoints;
+    unordered_map<int, u16string> watchRoots;
+    const int fdInotify;
```
wolfs: Should we use `RAII` for this and the next field as well?

src/test/groovy/net/rubygrapefruit/platform/file/LinuxFileEventFunctionsTest.groovy
lptr: Hah, good point, removed.

src/test/groovy/net/rubygrapefruit/platform/file/AbstractFileEventsTest.groovy
lptr: Added.

src/file-events/headers/linux_fsnotifier.h
lptr: This is in `linux_fsnotifier.h`. Do you mean to `linux_fsnotifier.cpp`?

wolfs: D'oh, yes I do.

lptr: What would be the advantage?

wolfs: Fewer files to look at and less indirection.

src/file-events/headers/linux_fsnotifier.h
lptr: We definitely could, and there are a bunch of other things we could handle similarly (like the Windows `HANDLE`s etc.). However, I think there are lower hanging fruits for now if we want to invest in refactoring.

wolfs: 👍

src/file-events/headers/linux_fsnotifier.h
lptr: You can't decrease the indirection, as the forward declarations would still be needed, making the `.cpp` file even longer. It also seems idiomatic to have a header file, though I don't observe this with the really internal types.

src/file-events/headers/linux_fsnotifier.h
lptr: OK, you convinced me, commit is in.

src/file-events/headers/linux_fsnotifier.h
lptr: And it's a nice place to get an overview of what's there. Think of it as our separation between interface and the `Default*` implementation.

src/file-events/headers/linux_fsnotifier.h
```diff
@@ -0,0 +1,69 @@
...
+
+using namespace std;
+
+class Server;
+
+struct Inotify {
+    Inotify();
+    ~Inotify();
+
+    const int fd;
```
wolfs: After this change it seems that `fd` should be private and the necessary methods should be added to `Inotify` so it can be. WYDT? As always, this can also be a follow up PR.

lptr: I tried that before, but we still refer to it when polling. I don't see how we can easily make that go away. Note that the field is a `const`, therefore only read access is allowed to it, which is not as bad as if this was a mutable field.

lptr: I think working with C++ and more importantly working with low-level C-style APIs allows two approaches: either model everything as an object, or accept that some primitives will be exposed. The extreme of the latter is to have everything as public `int`s, the extreme of the former is to reimplement the CLR. The current code seems to strike an acceptable balance for me.

wolfs:
@lptr You didn't merge the PR into `master`, but into `lptr/shared-threading`. Please merge/rebase it onto `master`.

PR #89, opened 2020-02-11
Rewrite macOS watcher using C++

lptr:
This is really just about code organization at this point. We still need better error handling and potentially use of std lib.


src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -1,116 +1,145 @@
...
-    JNIEnv *env;
-    jobject watcherCallback;
     CFRunLoopRef threadLoop;
-} watch_details_t;
+    bool invalidStateDetected;
+};
+
+Server::Server(JavaVM *jvm, JNIEnv *env, jobject watcherCallback, CFMutableArrayRef rootsToWatch, long latencyInMillis) {
+    this->jvm = jvm;
+    this->watcherCallback = env->NewGlobalRef(watcherCallback);
```
wolfs: Do we need to check here if `this->watcherCallback` is `NULL`?

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -1,116 +1,145 @@
...
-    // TODO What does this do?
-    size_t len = 0;
-    if (path != NULL) {
-        len = strlen(path);
-        for (char *p = path; *p != '\0'; p++) {
-            if (*p == '\n') {
-                *p = '\0';
-            }
-        }
+    if (pthread_create(&(watcherThread), NULL, EventProcessingThread, this) != 0) {
```
```suggestion
    if (pthread_create(&watcherThread, NULL, EventProcessingThread, this) != 0) {
```
Nit.

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -1,116 +1,145 @@
...
+        void *clientCallBackInfo,
+        size_t numEvents,
+        void *eventPaths,
+        const FSEventStreamEventFlags eventFlags[],
+        const FSEventStreamEventId eventIds[]);
+
+    void handleEvent(JNIEnv *env, char* path, FSEventStreamEventFlags flags);
+
+    // TODO: Move this to somewhere else
+    JNIEnv* getThreadEnv();
```
wolfs: Why doesn't server keep a reference to the `JNIEnv` associated with the thread?

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -120,138 +149,162 @@ static void *EventProcessingThread(void *data) {
...
+        log_warning(env, "Unknown event 0x%x for %s", flags, path);
+        type = FILE_EVENT_UNKNOWN;
     }
 
-    if (watcherStream != NULL) {
-        FSEventStreamStop(watcherStream);
-        FSEventStreamInvalidate(watcherStream);
-        FSEventStreamRelease(watcherStream);
-        // TODO: consider using FSEventStreamFlushSync to flush all pending events.
+    // TODO What does this do?
```
wolfs: Looks like it selects the first line from a multiline string by adding the terminator where the first line break is.

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -1,116 +1,145 @@
...
-
-/*
- * Apple specific functions.
- */
 #if defined(__APPLE__)
 
 #include "net_rubygrapefruit_platform_internal_jni_OsxFileEventFunctions.h"
 #include "generic.h"
 #include <CoreServices/CoreServices.h>
 #include <pthread.h>
```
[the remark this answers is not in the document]
lptr: Yes, that's the plan soon. Though using `std` on macOS might not be trivial, so I don't want to do it now.

src/main/cpp/apple_fsnotifier.cpp
lptr: I'll add a note to check this as well once we add proper error handling.

src/main/cpp/apple_fsnotifier.cpp
lptr: It could, though then we'd need to duplicate the error handling in case the `env` is not set.

src/main/cpp/apple_fsnotifier.cpp
lptr: Right. I guess the comment should have said "Why is this necessary?". Any ideas for that? We might as well remove this code and try to run the tests. I'm planning to do it in a separate PR.

wolfs: 👍

lptr:
My plan is to add error handling in a follow-up PR, yes.

src/main/cpp/apple_fsnotifier.cpp
wolfs: Currently, the thread crashes since `getThreadEnv()` will return `NULL`. I wouldn't call that error handling...

lptr: Yes, but we'll eventually have error handling implemented inside `getThreadEnv()`—which would then need to be duplicated. I don't want to do that.

Besides, even now before the thread crashes we log the fact that we couldn't get the `env` from the thread to `stderr`.

The `env` doesn't belong to the server as such, instead it belongs to the thread, and it is stored in a thread local there already. I'd argue that getting it from the thread local is the best way we have to do this.

PR #90, opened 2020-02-12
Pass if retrying the test makes it pass

wolfs:
instead of ignoring tests.


lptr:
This still keeps the build failing. I'll take a look at the underlying problem instead: https://github.com/gradle/native-platform/issues/91.

wolfs:
@lptr With the last commit ignoring test failures on Teamcity (https://github.com/gradle/native-platform/pull/90/commits/2e0d97d86f21024e7652ada7d5baf878dc092d64), it should pass now.

lptr:
Hm, still not fixed. In any case https://github.com/gradle/native-platform/pull/92 should fix the underlying problems, so let's not do this.

PR #94, opened 2020-02-17
Exception handling for macOS

lptr:
Fixes when merged via `gradle/gradle`: https://github.com/gradle/gradle/issues/12219

### DIFF
--- a/src/main/cpp/apple_fsnotifier.cpp
+++ b/src/main/cpp/apple_fsnotifier.cpp
@@ -7,6 +7,20 @@
 
 using namespace std;
 
+struct FileWatcherException : public exception {
+public:
+    FileWatcherException(const char *message) {
+        this->message = message;
+    }
+
+    const char * what () const throw () {
+        return message;
+    }
+
+private:
+    const char *message;
+};
+
 class Server;
 
 static void handleEventsCallback(
@@ -57,8 +71,7 @@ Server::Server(JNIEnv *env, jobject watcherCallback, CFArrayRef rootsToWatch, lo
     JavaVM* jvm;
     int jvmStatus = env->GetJavaVM(&jvm);
     if (jvmStatus < 0) {
-        log_severe(env, "Could not store jvm instance", NULL);
-        return;
+        throw FileWatcherException("Could not store jvm instance");
     }
 
     this->jvm = jvm;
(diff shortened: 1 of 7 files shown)


src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -222,56 +226,65 @@ JNIEnv* Server::getThreadEnv() {
...
     for (int i = 0; i < count; i++) {
         jstring path = (jstring) env->GetObjectArrayElement(paths, i);
-        char* watchedPath = java_to_char(env, path, result);
-        log_fine(env, "Watching %s", watchedPath);
+        char* watchedPath = java_to_char(env, path, NULL);
         if (watchedPath == NULL) {
-            log_severe(env, "Could not allocate string to store root to watch.", NULL);
-            // TODO Free resources
-            return NULL;
+            throw FileWatcherException("Could not allocate string to store root to watch");
```
wolfs: Don't we need to free resources when the exception is thrown?

lptr: That part is not done yet. I am actively learning about `unique_ptr` and friends to make sure they get released.

lptr:
@wolfs Yeah, I'm also unsure as to how to test error handling. :( Let me know if you have ideas. Meanwhile PTAL.

src/main/cpp/apple_fsnotifier.cpp
lptr: I think we are actually much better now wrt freeing resources when a problem occurs.

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -7,6 +7,20 @@
 
 using namespace std;
 
+struct FileWatcherException : public exception {
```
wolfs: I suppose this should move to `generic.cpp` at some point, right?

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -7,6 +7,20 @@
...
+public:
+    FileWatcherException(const char *message) {
+        this->message = message;
+    }
+
+    const char * what () const throw () {
```
```suggestion
    const char* what() const throw() {
```

Nit. What code style shall we use here? Can we maybe configure VSCode to autoformat the native code?

src/main/cpp/apple_fsnotifier.cpp
```diff
@@ -85,39 +103,39 @@ Server::Server(JNIEnv *env, jobject watcherCallback, CFArrayRef rootsToWatch, lo
...
-        log_severe(env, "Could not create FSEventStreamCreate to track changes", NULL);
-        // TODO Error handling
-        return;
+        throw FileWatcherException("Could not create FSEventStreamCreate to track changes");
     }
     this->watcherStream = watcherStream;
+
+    unique_lock<mutex> lock(watcherThreadMutex);
     this->watcherThread = thread(&Server::run, this);
+    this->watcherThreadStarted.wait(lock);
```
wolfs: Maybe we shouldn't wait forever here. Is that possible?

PR #99, opened 2020-02-18
Reorganize file watcher code

lptr:


src/main/headers/apple_fsnotifier.h
```diff
@@ -0,0 +1,61 @@
+#if defined(__APPLE__)
```
wolfs: All the new headers miss something like
```
#ifndef __INCLUDE_APPLE_FSNOTIFIER_H__
#define __INCLUDE_APPLE_FSNOTIFIER_H__
...
#endif
```

So you can include them twice and they only will be included once.

lptr: I ended up adding `#pragma once` instead.

PR #107, opened 2020-02-20
Register each watched path individually on macOS

lptr:
This is to make it work more similarly to the Windows watcher. Doing this will make it easier to start and stop watching things dynamically.

### DIFF
--- a/src/file-events/cpp/apple_fsnotifier.cpp
+++ b/src/file-events/cpp/apple_fsnotifier.cpp
@@ -27,7 +27,7 @@ WatchPoint::WatchPoint(Server* server, CFRunLoopRef runLoop, const u16string& pa
     }
     CFMutableArrayRef pathArray = CFArrayCreateMutable(NULL, 1, NULL);
     if (pathArray == NULL) {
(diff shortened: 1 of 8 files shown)


src/file-events/cpp/apple_fsnotifier.cpp
```diff
@@ -20,7 +20,13 @@ struct deletable_facet : Facet {
     }
 };
 
-EventStream::EventStream(Server* server, CFRunLoopRef runLoop, CFArrayRef rootsToWatch, long latencyInMillis) {
+EventStream::EventStream(Server* server, CFRunLoopRef runLoop, CFStringRef path, long latencyInMillis) {
+    CFMutableArrayRef pathArray = CFArrayCreateMutable(NULL, 1, NULL);
+    if (pathArray == NULL) {
+        throw FileWatcherException("Could not allocate array to store roots to watch");
```
wolfs:
```suggestion
        throw FileWatcherException("Could not allocate array to store root to watch");
```

lptr:
I measured 500 directories and it takes 18 ms to register a single `FSEventStream` and it takes 46 ms to register 500 individual ones.

PR #112, opened 2020-02-21
Some improvements to file events

lptr:


build.gradle
```diff
@@ -255,6 +255,33 @@ model {
...
+                    cppCompiler.args "-g"                       // Produce debug output
+                    cppCompiler.args "-pthread"                 // Force nicer threading
+                    cppCompiler.args "-pedantic"                // Disable non-standard things
+                    cppCompiler.args "--std=c++11"              // Enable C++11
+                    cppCompiler.args "-Wall"                    // All warnings
+                    cppCompiler.args "-Wextra"                  // Plus extra
+                    cppCompiler.args "-Wformat=2"               // Check printf format strings
+                    cppCompiler.args "-Werror"                  // Warnings are errors
+                    cppCompiler.args "-Wno-format-nonliteral"   // Allow printf to have dynamic format string
+                    linker.args "-pthread"
```
wolfs: Why do we now need to use `-pthread` which we didn't do before?

build.gradle
```diff
@@ -255,6 +255,33 @@ model {
...
+                } else if (targetPlatform.operatingSystem.windows) {
+                    cppCompiler.args "/DEBUG"                   // Produce debug output
+                    cppCompiler.args "/std:c++17"               // Won't hurt
```
wolfs: I think we should use the same `c++` version for Windows and Unix OSes.

build.gradle
lptr: Thad said we should. According to the crash logs we've always used pthreads, but he said in some weird cases the compiler could decide otherwise, and we don't want that.

build.gradle
lptr: You can't, it's C++14 by default on Windows, and you can't do C++11. We don't want to go above C++11 with Linux or macOS, though.

PR #115, opened 2020-02-24
Start/stop watching individual paths

lptr:


lptr:
@wolfs I'll merge this now, but please take a look when you are back.

src/file-events/headers/apple_fsnotifier.h
```diff
@@ -32,10 +32,13 @@ class WatchPoint {
...
 public:
-    Server(JNIEnv* env, jobject watcherCallback);
+    Server(JNIEnv* env, jobject watcherCallback, long latencyInMillis);
     ~Server();
 
-    void startWatching(const u16string& path, long latencyInMillis);
+    void startWatching(const u16string& path);
+    void stopWatching(const u16string& path);
+
+    // TODO This should be private
```
wolfs: Can we make it private? And maybe use `friend` for the static method which calls it?

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
```diff
@@ -60,16 +59,11 @@
...
      */
     // TODO How to set kFSEventStreamCreateFlagNoDefer when latency is non-zero?
-    public FileWatcher startWatching(Collection<String> paths, final long latency, final TimeUnit unit, FileWatcherCallback callback) {
-        return createWatcher(paths, callback, new WatcherFactory() {
-            @Override
-            public FileWatcher createWatcher(String[] canonicalPaths, NativeFileWatcherCallback callback) {
-                return startWatching(canonicalPaths, unit.toMillis(latency), callback);
-            }
-        });
+    public FileWatcher startWatcher(long latency, TimeUnit unit, FileWatcherCallback callback) {
```
wolfs: Should this be called `startWatcher`? We start the thread there, though no watching actually starts.

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -8,14 +10,25 @@ using namespace std;
...
     this->buffer = (FILE_NOTIFY_INFORMATION*) malloc(EVENT_BUFFER_SIZE);
     ZeroMemory(&this->overlapped, sizeof(OVERLAPPED));
     this->overlapped.hEvent = this;
     this->directoryHandle = directoryHandle;
     this->status = WATCH_UNINITIALIZED;
+    unique_lock<mutex> lock(listenerMutex);
+    QueueUserAPC(startWatchCallback, serverThreadHandle, (ULONG_PTR) this);
+    listenerStarted.wait(lock);
+    if (status != WATCH_LISTENING) {
+        throw new FileWatcherException("Couldn't start listening");
```
wolfs: Can we add the path we try to watch to this exception?

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -182,25 +183,18 @@ void Server::startWatching(JNIEnv* env, const u16string& path) {
...
-        case WATCH_LISTENING:
-            watchPoints.push_back(watchPoint);
-            break;
-        default:
-            log_severe(env, "Couldn't start listening to '%ls': %d", pathW.c_str(), ret);
-            delete watchPoint;
-            // TODO Error handling
-            break;
-    }
+    watchPoints.emplace(piecewise_construct,
```
wolfs: Should we also check here if the location is already watched?

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -182,25 +183,18 @@ void Server::startWatching(JNIEnv* env, const u16string& path) {
...
+        forward_as_tuple(path),
+        forward_as_tuple(this, path, directoryHandle, threadHandle));
 }
 
-void Server::reportFinished(WatchPoint* watchPoint) {
-    watchPoints.remove(watchPoint);
-    delete watchPoint;
+void Server::stopWatching(JNIEnv* env, const u16string& path) {
+    watchPoints.find(path)->second.close();
```
wolfs: Should we check here if we are watching the path?

src/test/groovy/net/rubygrapefruit/platform/file/AbstractFileEventsTest.groovy
```diff
@@ -280,6 +280,21 @@ abstract class AbstractFileEventsTest extends Specification {
         expectedChanges.await()
     }
 
+    def "does not receive events after directory is unwatched"() {
```
wolfs: There are tests missing for what happens when
- `startWatching` twice for the same path
- `stopWatching` for a non-watched path
- `stopWatching` twice for the same watched path.

src/file-events/cpp/apple_fsnotifier.cpp
```diff
@@ -190,29 +191,27 @@ void Server::handleEvent(JNIEnv* env, char* path, FSEventStreamEventFlags flags)
     reportChange(env, type, pathStr);
 }
 
-void Server::startWatching(const u16string& path, long latencyInMillis) {
-    watchPoints.emplace_back(this, threadLoop, path, latencyInMillis);
+void Server::startWatching(const u16string& path) {
+    if (watchPoints.find(path) != watchPoints.end()) {
+        throw new FileWatcherException("Already watching path");
+    }
+    // TODO Is this necessary
```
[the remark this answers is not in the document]
lptr: I just checked with Thad, and it seems to be. I'm removing the TODOs soon.

src/file-events/cpp/apple_fsnotifier.cpp
```diff
@@ -190,29 +191,27 @@ void Server::handleEvent(JNIEnv* env, char* path, FSEventStreamEventFlags flags)
     reportChange(env, type, pathStr);
 }
 
-void Server::startWatching(const u16string& path, long latencyInMillis) {
-    watchPoints.emplace_back(this, threadLoop, path, latencyInMillis);
+void Server::startWatching(const u16string& path) {
+    if (watchPoints.find(path) != watchPoints.end()) {
+        throw new FileWatcherException("Already watching path");
```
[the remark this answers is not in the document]
lptr: I think we should opt for strictness in the beginning. If it becomes troublesome we can make things more lenient, but I'm not sure what would convince me that we need to do just that.

The Java side should be aware of all the watchers that are registered, and it should signal a problem if we tried to register the same watcher twice, so I think it'll be fine to fail always.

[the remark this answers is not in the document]
lptr: I made a not to add Javadocs.

src/file-events/headers/apple_fsnotifier.h
lptr: We will, yeah.

src/main/java/net/rubygrapefruit/platform/internal/jni/OsxFileEventFunctions.java
lptr: That's a good question. It's a bit painful to rename things around here, so I didn't play around with it yet, but I definitely want to clarify the terminology once we have Linux done as well.

src/file-events/cpp/win_fsnotifier.cpp
lptr: No. Dynamic strings are not a thing in C++ exceptions unfortunately. :/

wolfs: :(

src/file-events/cpp/win_fsnotifier.cpp
lptr: Good point. Fixing it.

src/file-events/cpp/win_fsnotifier.cpp
lptr: Yes, fixing.

src/test/groovy/net/rubygrapefruit/platform/file/AbstractFileEventsTest.groovy
lptr: Good point, adding.

PR #117, opened 2020-02-26
Various improvements

lptr:


lptr:
@wolfs These are some recent improvements we made with Thad and also some that I did in response to your post-merge reviews on previous PRs. I'll merge them now, please take a look when you have time.

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -77,13 +86,6 @@ void WatchPoint::listen() {
 void WatchPoint::handleEvent(DWORD errorCode, DWORD bytesTransferred) {
```
wolfs: Let's remove the `errorCode` argument.

lptr: Done as part of https://github.com/gradle/native-platform/pull/119.

PR #120, opened 2020-02-28
Shared threading for file watchers

lptr:
Share a command pattern-based model for offloading work to the processing thread between the two existing implementations (macOS and Windows).

Doing so separates the complexity of thread handling from the complexity of the individual OS APIs.

Note: we should probably further simplify things by extracting the following methods into a separate pure virtual class, and make `AbstractServer` into a concrete class instead:

- `void registerPath(const u16string& path)`
- `void unregisterPath(const u16string& path)`
- `void terminate()`

...essentially moving all the stuff that happens on the background thread to a separate class.

This could probably allow us to move most of the `~Server()` dtors into `AbstractServer`, and to remove `startThread()`.

However, for now we'll stop here, and instead reimplement the Linux watcher on top of this architecture as a next step.


build.gradle
```diff
@@ -279,6 +279,8 @@ model {
                     cppCompiler.args "/Zc:throwingNew"          // Assume new throws on error
                     cppCompiler.args "/W3"                      // Enable lots of warnings, disbale individual warnings with /WD<NUM>
                     cppCompiler.args "/WX"                      // Warnings are errors
+                    cppCompiler.args "/D_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING"
```
wolfs: Is that necessary or did the problem only occur on Windows 24?

src/file-events/headers/generic_fsnotifier.h
```diff
@@ -36,13 +39,64 @@ struct FileWatcherException : public exception {
...
 };
 
+class AbstractServer;
+
+class Command {
+public:
+    Command(){};
+    virtual ~Command(){};
+
+    void execute(AbstractServer* server) {
```
wolfs: Can we reduce the scope and not pass all of `AbstractServer` here? It seems like a better idea to me to move everything which is used on the thread to a separate object, so the command can't access/call things which it isn't suppose to call. Moreover, the comments that some methods only should be used on the thread can be removed by moving those methods to the new class - then it is only possible to call the methods from the threads.

WDYT?

build.gradle
lptr: This is a warning with C++17, so it's not related to the problem we've seen with the `windows24` machine.

src/file-events/headers/generic_fsnotifier.h
lptr: I fully agree. I wrote about this in the issue description and why I want to stop for now with the structural improvements. But what you describe is the endgame, yes.

Note that any kind of refactoring is pretty complicated, as VSCode doesn't refactor the other platforms, and in general things need to be checked on every platform.

We'll get there, but I think for now we're good.

wolfs: 👍

PR #121, opened 2020-03-03
Add tests for created/deleted directories

lptr:


src/test/groovy/net/rubygrapefruit/platform/file/AbstractFileEventsTest.groovy
```diff
@@ -676,7 +711,7 @@ abstract class AbstractFileEventsTest extends Specification {
         void pathChanged(Type type, String path) {
             def canonicalFile
             try {
-                canonicalFile = new File(path).canonicalFile
+                canonicalFile = new File(path)
             } catch (IOException e) {
                 throw new RuntimeException("Couldn't canonicalize path: '$path'", e)
```
wolfs: Do we need all this `try ... catch` any more now that we are not canonicalizing anymore?

lptr: Fixed!

PR #122, opened 2020-03-03
Handle watched directory deleted on Linux

lptr:


src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -155,21 +166,29 @@ void Server::processCommandsOnThread() {
...
+        ? ""
+        : event->name;
+    log_fine(env, "Event mask: 0x%x for %s (wd = %d, cookie = 0x%x, len = %d)", mask, eventName, event->wd, event->cookie, event->len);
     if (IS_ANY_SET(mask, IN_UNMOUNT)) {
         return;
     }
-    // TODO Do we need error handling here?
+
     u16string path = watchRoots[event->wd];
+    if (path.empty()) {
```
wolfs: So if an element is not part of the map, then an empty String is returned? Because it uses the empty constructor of `u16string`?

lptr: That's my interpretation, too, and it matches what I've seen. It's weird.

PR #123, opened 2020-03-03
Handle exception while processing operating system events

lptr:


src/file-events/cpp/generic_fsnotifier.cpp
```diff
@@ -163,9 +164,55 @@ void AbstractServer::reportChange(JNIEnv* env, int type, const u16string& path)
...
     env->DeleteLocalRef(javaPath);
+
+    jthrowable exception = env->ExceptionOccurred();
+    if (exception != nullptr) {
+        env->ExceptionDescribe();
+        env->ExceptionClear();
+
+        jclass exceptionClass = env->GetObjectClass(exception);
+        assert(exceptionClass != nullptr);
+        jmethodID getMessage = env->GetMethodID(exceptionClass, "getMessage", "()Ljava/lang/String;");
```
wolfs: Do you need to delete the `getMessage` local ref as well?

src/file-events/cpp/generic_fsnotifier.cpp
```diff
@@ -163,9 +164,55 @@ void AbstractServer::reportChange(JNIEnv* env, int type, const u16string& path)
...
+        assert(getMessage != nullptr);
+        jstring javaMessage = (jstring) env->CallObjectMethod(exception, getMessage);
+        assert(javaMessage != nullptr);
+        string message = javaToUtf8String(env, javaMessage);
+        env->DeleteLocalRef(javaMessage);
+
+        jclass classClass = env->FindClass("java/lang/Class");
+        assert(classClass != nullptr);
+        jmethodID getClassName = env->GetMethodID(classClass, "getName", "()Ljava/lang/String;");
```
wolfs: Do you need to delete the `getClassName ` local ref as well?

src/file-events/cpp/generic_fsnotifier.cpp
```diff
@@ -163,9 +164,55 @@ void AbstractServer::reportChange(JNIEnv* env, int type, const u16string& path)
     jstring javaPath = env->NewString((jchar*) path.c_str(), (jsize) path.length());
     env->CallVoidMethod(watcherCallback, watcherCallbackMethod, type, javaPath);
     env->DeleteLocalRef(javaPath);
+
+    jthrowable exception = env->ExceptionOccurred();
+    if (exception != nullptr) {
+        env->ExceptionDescribe();
```
wolfs: Shouldn't we better hand the exception to the logging instead of printing the stacktrace to stderr?

src/file-events/cpp/generic_fsnotifier.cpp
```diff
@@ -163,9 +164,55 @@ void AbstractServer::reportChange(JNIEnv* env, int type, const u16string& path)
...
+void AbstractServer::reportError(JNIEnv* env, const exception& exception) {
+    jclass exceptionClass = env->FindClass("net/rubygrapefruit/platform/NativeException");
+    assert(exceptionClass != nullptr);
+    u16string message = utf8ToUtf16String(exception.what());
+    jstring javaMessage = env->NewString((jchar*) message.c_str(), (jsize) message.length());
+    jmethodID constructor = env->GetMethodID(exceptionClass, "<init>", "(Ljava/lang/String;)V");
+    jobject javaException = env->NewObject(exceptionClass, constructor, javaMessage);
+    assert(javaException != nullptr);
+    env->CallVoidMethod(watcherCallback, watcherReportErrorMethod, javaException);
+    env->DeleteLocalRef(exceptionClass);
```
wolfs: Do you also need to delete the other local refs to `javaMessage`, `constructor` and `javaException`?

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -70,28 +74,46 @@ void WatchPoint::listen() {
...
+    const u16string& path = watchPoint->path;
+
+    try {
+        switch (errorCode) {
+            case ERROR_SUCCESS:
                 break;
+            case ERROR_OPERATION_ABORTED:
+                log_fine(env, "Finished watching '%s'", utf16ToUtf8String(path).c_str());
+                reportFinished(path);
+                return;
```
wolfs:
```suggestion
                break;
```

Why not use `break` here as well?

src/file-events/cpp/generic_fsnotifier.cpp
lptr: I don't think these are terribly important in any case, but the example I copied from didn't explicitly delete `jmethodID`s.

src/file-events/cpp/generic_fsnotifier.cpp
lptr: We could, though we don't currently have a way to log exceptions from the native side. Note that we are currently passing an already formatted string to logging, and here we'd need to pass a `jobject`.

I think for now this will do.

wolfs: My fear is that the logging will be lost instead of being part of the daemon log. We saw this when printing to stdout as well. Maybe it will work since it uses Java code to print of the stacktrace?

lptr: Actually, I can do this in `NativeFileWatcherCallback` already...

src/file-events/cpp/win_fsnotifier.cpp
lptr: Because this is the point where we want to stop listening to this watch point. Using `break` here would result in the watcher being re-established.

wolfs: Ah, now I got it. For some reason I was under the impression that the `switch` statement is the last statement in the block. Maybe we could extract the rest of the code here into a separate method and call it from the `case ERROR_SUCCESS` branch? Then `break` would work fine.

lptr: Yes, but that would add another variation to this method, maybe `handleEventsWithoutErrorCode()`. The extra indirection is not making the code more readable for me.

This `switch` at the top simply sorts out the cases we can handle early, and lets `ERROR_SUCCESS` pass through.

If you feel strongly I can try rearranging things a bit, but I don't think the end result will be any more readable.

wolfs: We can leave it like it is.

src/file-events/cpp/generic_fsnotifier.cpp
wolfs: Given that we won't throw exceptions from the callback all the time, it should be fine to leave it as is. For me, it would be less cognitive overhead if all the `j*` typed local variables would be deleted, so I don't need to remember which ones to frees and which not when reading the code.

lptr: I just checked, and you can't pass a `jmethodID` to `DeleteLocalRef()` because it expects a `jobject`.

src/file-events/cpp/generic_fsnotifier.cpp
lptr: Ah, I missed a couple things here, fix: https://github.com/gradle/native-platform/pull/126.

PR #131, opened 2020-03-10
Fix hangs and crashes on Windows and Linux

lptr:


lptr:
Some logs of Windows builds failing when starting/stopping watchers while changes are coming in: https://builds.gradle.org/buildConfiguration/GradleNative_NativePlatform_BuildWindows/32433906?buildTab=tests&focusLine=0&showLog=32433906_926_512.808.839.921

lptr:
I cannot seem to reproduce this failure on Windows:

https://e.grdev.net/s/pkww53merbvf2/tests/si6uhuz4tdufy-3ktq35w3rgc7w

```text
18:26:31.128  INFO    Making change #7...
18:26:31.128  FINE    Change detected: 0x1 'file7.txt'
18:26:31.128  INFO    Received: CREATED - C:\tcagent1\temp\buildTmp\junit941827092003971320\can start and stop watching directory while changes are being made to its contents\root\file7.txt
18:26:31.147  INFO    Making change #8...
18:26:31.147  INFO    Tearing down
18:26:31.148  FINE    Change detected: 0x1 'file8.txt'
18:26:31.148  INFO    Received: CREATED - C:\tcagent1\temp\buildTmp\junit941827092003971320\can start and stop watching directory while changes are being made to its contents\root\file8.txt
18:26:31.148  INFO    Detected overflow for 
18:26:31.148  INFO    Received: INVALIDATE - 
net.rubygrapefruit.platform.NativeException: Couldn't start watching, error = 998: 
18:26:31.156  FINE    Path is not watched: 
18:26:31.156  FINE    Stopping thread
```

Looks like `change #8` is received at the same time as we are tearing down the watcher `Server`.

All of these log messages refer to an empty path:

```text
18:26:31.148  INFO    Detected overflow for 
18:26:31.148  INFO    Received: INVALIDATE - 
net.rubygrapefruit.platform.NativeException: Couldn't start watching, error = 998: 
18:26:31.156  FINE    Path is not watched: 
```

My hunch is that the corresponding `WatchPoint` has been destructed, and as such has its path freed up, which means it becomes an empty string. I wish I had a more reliable understanding of how this works in C++.

lptr:
OMG, okay, so this is valid C++:

```cpp
    u16string* thing = new u16string(u"alma");
    fprintf(stderr, ">>>> length: %d\n", (int) thing->length());
    fprintf(stderr, ">>>> string: %s\n", utf16ToUtf8String(*thing).c_str());
    delete thing;
    fprintf(stderr, ">>>> length 2: %d\n", (int) thing->length());
    fprintf(stderr, ">>>> string 2: %s\n", utf16ToUtf8String(*thing).c_str());
```

...and it prints this:

```text
>>>> length: 4
>>>> string: alma
>>>> length 2: 0
>>>> string 2:
```

lptr:
I cannot express how horrified I am about this. My `WatchPoint` object has a `const u16string path`, and that `const` thing just simply changes its value to become an empty string, but otherwise stays available with absolutely no sing of a problem. Gah!

lptr:
And this explains the crashes, too: we try to read the events into the buffer that has also been freed up, causing:

```text
ERROR_NOACCESS

998 (0x3E6)

Invalid access to memory location.
```

lptr:
This is how we free up things on Windows now:

1) remove path from `watchPoints`
2) leads to `WatchPoint::~WatchPoint()` being called
3) calls `CancelIO()` which schedules `ERROR_OPERATION_ABORTED`
4) handles pending events in `SleepEx()`
5) destructs `path` and `buffer`

I think it is possible that Windows reports events (and calls the callback for) after 5). In this case we end up handling these events using a destructed `WatchPoint` that an invalid `path` and `buffer`.

Maybe it means that in some circumstances in 4) we don't actually process all queued events for the watch point? What are those circumstances?

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -231,9 +241,46 @@ Server::~Server() {
...
+                break;
+            default:
+                break;
+        }
+    }
+
+    // If there are any pending watchers, wait for them to finish
+    if (pendingWatchPoints > 0) {
+        log_fine(env, "Waiting for %d pending watch points to finish", pendingWatchPoints);
+        SleepEx(SERVER_CLOSE_TIMEOUT_IN_MS, true);
```
wolfs: Same here, I don't think one round is necessarily enough to receive all the `ERROR_OPERATION_ABORTED` events.

wolfs:
We still have crashes on Windows. I think this comes from the problem with stopping multiple watch points which is exposed by the native tests. Let me come up with a test.

src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -22,16 +22,19 @@ static int registerWatchPoint(const u16string& path, shared_ptr<Inotify> inotify
...
 WatchPoint::WatchPoint(const u16string& path, shared_ptr<Inotify> inotify)
-    : watchDescriptor(registerWatchPoint(path, inotify))
-    , inotify(inotify) {
+    : status(LISTENING)
+    , watchDescriptor(registerWatchPoint(path, inotify))
+    , inotify(inotify)
+    , path(path) {
 }
 
 WatchPoint::~WatchPoint() {
```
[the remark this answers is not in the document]
lptr: Let's do code beautification later. I have moved things in and out of that destructor a lot recently, I'm not sure it's going to be needed or not later on.

src/file-events/cpp/win_fsnotifier.cpp
lptr: This is actually fine, as all the `CancelIoEx()` happen on the same thread, right before calling `SleepEx()`. So all the events are handled in one go, provided there is enough time in the timeout (for which a whole second should do).

lptr:
I'm not sure what the crashes on Windows were caused by, but now we are more robust in how we cancel watch points.

Looks like `CancelIo()` was happy to return `true` even when it later didn't produce an `ERROR_OPERATION_ABORTED` event. Now we use `CancelIoEx()` that actually returns `false` in some cases that seem to be exactly the cases where we used to miss the `ERROR_OPERATION_ABORTED` event. We now handle that case and as a result it looks like we can account for all the watch points.

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -197,7 +207,7 @@ void Server::handleEvent(JNIEnv* env, const u16string& path, FILE_NOTIFY_INFORMA
         }
     }
 
-    log_fine(env, "Change detected: 0x%x '%s'", info->Action, utf16ToUtf8String(changedPath).c_str());
+    // log_fine(env, "Change detected: 0x%x '%s'", info->Action, utf16ToUtf8String(changedPath).c_str());
```
wolfs: Is this commented out accidentally?

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -42,9 +42,17 @@ void WatchPoint::cancel() {
...
+        BOOL ret = CancelIoEx(directoryHandle, &overlapped);
         if (!ret) {
-            throw FileWatcherException("Couldn't cancel I/O", path, GetLastError());
+            status = FINISHED;
+            DWORD lastError = GetLastError();
+            if (lastError == ERROR_NOT_FOUND) {
+                // Do nothing, looks like this is a typical scenario
+                log_fine(server->getThreadEnv(), "Watch point already finished %s", utf16ToUtf8String(path).c_str());
+            } else {
+                // TODO Should we throw here instead?
```
wolfs: We should do the same on Linux and Windows. Is this maybe different when we cancel watching as a request to unregister or when we cancel watching when we close the watcher?

src/test/groovy/net/rubygrapefruit/platform/file/AbstractFileEventsTest.groovy
```diff
@@ -336,7 +336,8 @@ abstract class AbstractFileEventsTest extends Specification {
 
     def "can start and stop watching directory while changes are being made to its contents"() {
         given:
-        def numberOfParallelWriters = 100
+        def numberOfParallelWritersPerWatchedDirectory = 100
```
wolfs: That is currently too high. Shall we lower it to 25? I can polish the test in another PR.

src/file-events/cpp/win_fsnotifier.cpp
lptr: No, but it shouldn't be merged. I'll add a TODO.

src/file-events/cpp/win_fsnotifier.cpp
lptr: I don't understand, sorry. What should we do the same?

If your question is about when do we get `ERROR_NOT_FOUND` – I only saw it only during the stress test, only while terminating the server, and even then only once or twice per execution.

wolfs: On Linux, we currently throw an exception when we can't cancel the watcher. On Windows, we would now only log the problem. I think it would make sense to do the same on both OSes. For example, only logging the problem is fine for me.

lptr: I see what you mean. I'm about to take a look at Linux now.

src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -218,28 +251,44 @@ void Server::handleEvent(JNIEnv* env, const inotify_event* event) {
...
+    }
+    return fdWatch;
+}
+
 void Server::registerPath(const u16string& path) {
-    if (watchPoints.find(path) != watchPoints.end()) {
+    auto it = watchPoints.find(path);
+    if (it != watchPoints.end()) {
+        auto& watchPoint = it->second;
+        if (watchPoint.status != FINISHED) {
```
wolfs: We probably should removed the `FINISHED` nodes at some point from the `watchPoints`, so we can free up the memory.

This is currently not a problem since we stop the watchers every build.

src/file-events/cpp/win_fsnotifier.cpp
wolfs: Let's make this throw on a separate PR.

src/file-events/cpp/win_fsnotifier.cpp
wolfs: OK. Let's handle it when we tackle better logging.

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -231,9 +282,48 @@ Server::~Server() {
...
     }
+
+    // We have received termination, cancel all watchers
+    log_fine(env, "Finished with run loop, now cancelling remaining watch points", NULL);
+    int pendingWatchPoints = 0;
+    for (auto& it : watchPoints) {
+        auto& watchPoint = it.second;
+        switch (watchPoint.status) {
+            case LISTENING:
+                if (watchPoint.cancel()) {
```
wolfs: We should handle the exceptions here, so we can close the remaining handles.

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -259,13 +353,12 @@ void Server::registerPath(const u16string& path) {
 void Server::unregisterPath(const u16string& path) {
     u16string longPath = path;
     convertToLongPathIfNeeded(longPath);
-    reportFinished(longPath);
-}
-
-void Server::reportFinished(const u16string path) {
-    if (watchPoints.erase(path) == 0) {
+    auto it = watchPoints.find(longPath);
+    if (it == watchPoints.end()) {
```
wolfs: We should check if the watch point is finished here if we find one for the path.

src/file-events/cpp/linux_fsnotifier.cpp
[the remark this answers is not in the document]
lptr: Removed in https://github.com/gradle/native-platform/pull/142

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -35,19 +36,16 @@ WatchPoint::WatchPoint(Server* server, const u16string& path)
 }
 
 WatchPoint::~WatchPoint() {
```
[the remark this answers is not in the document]
lptr: Removed in https://github.com/gradle/native-platform/pull/142

src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -76,33 +79,61 @@ Server::Server(JNIEnv* env, jobject watcherCallback)
...
+                break;
+            default:
+                break;
+        }
+    }
+
+    // If there are any pending watchers, wait for them to finish
+    if (pendingWatchPoints > 0) {
+        log_fine(env, "Waiting for %d pending watch points to finish", pendingWatchPoints);
+        processQueues(SERVER_CLOSE_TIMEOUT_IN_MS);
```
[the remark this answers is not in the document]
lptr: FTR, we are making an attempt here to close everything, though it's not really necessary, given that the `inotify` will be closed via its destructor in any case. Maybe we should remove this altogether.

src/file-events/cpp/win_fsnotifier.cpp
lptr: Seems like `0` time is also fine for this.

src/file-events/cpp/win_fsnotifier.cpp
lptr: Fixed in https://github.com/gradle/native-platform/pull/142

PR #132, opened 2020-03-11
Handle empty file name received in event notification on Windows

lptr:
It is unclear when we should receive an event with an empty file name (maybe when the watched directory itself is changed?), and why these things started to cause `StringIndexOutOfBoundsException` (or whether they are the culprit). However, this case needs to be handled, so here's a fix.

I'm also strengthening our tests by checking if the reported change is an absolute path and failing the test if it isn't.


lptr:
Merged.

PR #135, opened 2020-03-12
Fix hangs on Linux

wolfs:


wolfs:
Doesn't seem to be fixed completely: https://builds.gradle.org/buildConfiguration/GradleNative_NativePlatform_BuildLinuxAarch64/32483279?

It seems like we receive events even though the path is not registered, yet. Probably the watch point has already been deleted at that point...

src/test/groovy/net/rubygrapefruit/platform/file/AbstractFileEventsTest.groovy
```diff
@@ -382,12 +375,12 @@ abstract class AbstractFileEventsTest extends Specification {
             watcher.startWatching(rootDir)
             readyLatch.await()
             startModifyingLatch.countDown()
-            Thread.sleep(200)
+            Thread.sleep(500)
             watcher.close()
+            assert uncaughtFailureOnThread.empty
```
lptr: This is now asserted thrice, why?

wolfs: I am asserting it here, so the test ends early when the first error occurs. I removed the other assertion.

wolfs:
@lptr This PR fixes the hang and converts it into a crash - only much less frequently. Using a snapshot from this PR actually makes the Linux test pass on the "keep watches alive during build" PR: https://builds.gradle.org/buildConfiguration/Gradle_Check_VfsRetention_28_Trigger/32486344?buildTab=overview

I suggest we merge it and fix the remaining crashes on Windows and Linux separately. WDYT?

lptr:
OK. I'll merge it now sow I can rebase my fix for Windows on top. We can bump Gradle tomorrow and check how we can get rid of the remaining crash.

PR #138, opened 2020-03-12
Reorganize tests

lptr:


lptr:
I tried with subclasses as well, but it's ugly for other reasons. I think it's fine as it is.

FreeBSD is fixed now. Unfortunately `@Requires` is not inherited, and I couldn't find a better way to do this right now but to re-iterate the condition for each test instead.

PTAL.

lptr:
Which methods would you have moved to the fixture?

src/test/groovy/net/rubygrapefruit/platform/file/SymlinkFileEventFunctionsTest.groovy
```diff
@@ -0,0 +1,98 @@
...
+
+import net.rubygrapefruit.platform.NativeException
+import net.rubygrapefruit.platform.internal.Platform
+import spock.lang.Ignore
+import spock.lang.Requires
+
+import static java.nio.file.Files.createSymbolicLink
+import static net.rubygrapefruit.platform.file.FileWatcherCallback.Type.MODIFIED
+
+@Requires({ Platform.current().macOs || Platform.current().linux || Platform.current().windows })
```
wolfs: Does using `assumeTrue` in `setup` work instead of duplicating the class level `@Requires`?

lptr: Unfortunately it doesn't. I tried that and a number of other things. I think this is good enough.

PR #140, opened 2020-03-13
Watch multiple directories in stress test

wolfs:


wolfs:
We already have this test on `master` now.

PR #141, opened 2020-03-13
Fix crash when unwatching symlinked directory

wolfs:


src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -105,7 +105,11 @@ void Server::runLoop(JNIEnv* env, function<void(exception_ptr)> notifyStarted) {
         auto& watchPoint = it.second;
         switch (watchPoint.status) {
             case LISTENING:
-                watchPoint.cancel();
+                try {
+                    watchPoint.cancel();
+                } catch (const exception& ex) {
+                    log_warning(env, ex.what(), NULL);
```
lptr: This makes sure you don't crash even if the exception message contains something like `%s`:

```suggestion
                    log_warning(env, "%s", ex.what());
```

wolfs:
The changes have been included in #131.

PR #143, opened 2020-03-16
Wait for cancelled watch point to be finished on Linux

lptr:
Fixes a problem when we stop and start watching the same directory in quick succession on Linux.

### DIFF
--- a/src/file-events/cpp/linux_fsnotifier.cpp
+++ b/src/file-events/cpp/linux_fsnotifier.cpp
@@ -291,7 +291,7 @@ void Server::unregisterPath(const u16string& path) {
     watchPoint.cancel();
     processQueues(CLOSE_TIMEOUT_IN_MS);
     if (watchPoint.status != FINISHED) {
-        log_warning(getThreadEnv(), "Could not cancel watch point %s", utf16ToUtf8String(path).c_str());
+        throw FileWatcherException("Could not cancel watch point %s", path);
     } else {
         watchRoots.erase(watchPoint.watchDescriptor);
         watchPoints.erase(path);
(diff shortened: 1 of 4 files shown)


src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -289,6 +289,8 @@ void Server::unregisterPath(const u16string& path) {
     }
     auto& watchPoint = it->second;
     watchPoint.cancel();
+    // Make sure we handle the cancellation
```
wolfs:
```suggestion
    // Make sure the status of the watch point transitions to FINISHED
```

Or something along those lines would be better I think.

Can we be sure that the event already did arrive when we call `processQueues`?

lptr: I chose a different path. PTAL.

src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -289,6 +289,13 @@ void Server::unregisterPath(const u16string& path) {
     }
     auto& watchPoint = it->second;
     watchPoint.cancel();
+    processQueues(CLOSE_TIMEOUT_IN_MS);
+    if (watchPoint.status != FINISHED) {
+        log_warning(getThreadEnv(), "Could not cancel watch point %s", utf16ToUtf8String(path).c_str());
```
wolfs: So starting from here we are in a bad state, right? We can't register the watch point anymore and we don't know if it will ever transition to finished.

Does this actually happen?

lptr: What do you suggest, that we remove the watch point anyway?

wolfs: We could throw an exception right here. When we erase the watchpoint, then an event coming in later for the watch point would crash the JVM, right?

I would really prefer throwing an exception over crashing the JVM. Logging the warning is also fine for now I think.

lptr: We don't remove the watch point in this case, so any events coming in will not crash the JVM. Throwing here doesn't necessarily change that, since that only results in the Java side getting an exception—there's no guarantee the Java side will tear down or anything.

We can nevertheless throw here, too, so that the Java side gets some indication there was a problem.

lptr: Throwing now.

PR #144, opened 2020-03-17
Handle stop watching many directories when they have been deleted

wolfs:


lptr:
@wolfs PTAL. (Can't invite to review)

PR #146, opened 2020-03-18
Load `NativeException` class when we start the server

wolfs:
Under some circumstances, we seem to be unable to load the class when trying to close the server and fail when trying to report an exception to the Java side.

See https://stackoverflow.com/questions/20752352/classnotfoundexception-when-finding-a-class-in-jni-background-thread
and https://stackoverflow.com/questions/13263340/findclass-from-any-thread-in-android-jni
for reasons which might cause the problem.


wolfs:
Superseded by #148 and #149.

PR #147, opened 2020-03-18
Remove `assert()` calls for `NULL` checks

lptr:
Asserts crash the VM with a message we cannot track on the Java side. Instead, we'll let things proceed, and let the VM crash with `SIGSEGV`, which forces the JVM to produce a crash log. Which is a lot nicer.

### DIFF
--- a/src/file-events/cpp/generic_fsnotifier.cpp
+++ b/src/file-events/cpp/generic_fsnotifier.cpp
@@ -1,4 +1,3 @@
-#include <assert.h>
 #include <codecvt>
 #include <locale>
 #include <sstream>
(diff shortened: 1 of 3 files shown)


src/file-events/cpp/generic_fsnotifier.cpp
```diff
@@ -175,20 +175,14 @@ void AbstractServer::reportChange(JNIEnv* env, int type, const u16string& path)
         env->ExceptionClear();
 
         jclass exceptionClass = env->GetObjectClass(exception);
-        assert(exceptionClass != nullptr);
```
wolfs: Can we remove the `#include <assert.h>` statements as well?

lptr: Getting rid of them as well now.

PR #150, opened 2020-03-20
Logging, take 2

lptr:
A number of improvements around logging:

- no initialization of the logging required, instead we initialize in `JNI_OnLoad()`
- log level is queried from the native side, and cached for 1 second
- no need to pass `JNIEnv*` to logging anymore
- `log_fine()` etc. is replaced with `log(FINE, ...)`

- added `jni_support.cpp` with all the JNI stuff
- assed `JniSupport` for types that need to work with JNI
- added `JniGlobalRef` RAII

### DIFF
--- a/src/file-events/cpp/logging.cpp
+++ b/src/file-events/cpp/logging.cpp
@@ -1,7 +1,5 @@
 #include "logging.h"
 
-Logging* logging;
-
 Logging::Logging(JavaVM* jvm)
     : JniSupport(jvm)
     , clsLogger(getThreadEnv(), "net/rubygrapefruit/platform/internal/jni/NativeLogger")
--- a/src/file-events/cpp/logging.cpp
+++ b/src/file-events/cpp/logging.cpp
@@ -13,6 +13,10 @@ Logging::Logging(JNIEnv* env, int level)
     , logMethod(env->GetStaticMethodID(clsLogger, "log", "(ILjava/lang/String;)V")) {
     printlog(LOG_CONFIG, "Initialized logging to level %d\n", level);
 }
+Logging::~Logging() {
+    JNIEnv* env = getThreadEnv();
+    env->DeleteGlobalRef(clsLogger);
+}
 
 void Logging::printlog(int level, const char* fmt, ...) {
     if (minimumLogLevel > level) {
--- a/src/file-events/cpp/logging.cpp
+++ b/src/file-events/cpp/logging.cpp
@@ -1,3 +1,5 @@
+#include <iostream>
+
 #include "logging.h"
 
 Logging::Logging(JavaVM* jvm)
@@ -26,7 +28,7 @@ void Logging::send(LogLevel level, const char* fmt, ...) {
 
     JNIEnv* env = getThreadEnv();
     if (env == NULL) {
-        fprintf(stderr, "%s\n", buffer);
+        cerr << buffer << endl;
     } else {
         jstring logString = env->NewStringUTF(buffer);
         env->CallStaticVoidMethod(clsLogger.get(), logMethod, level, logString);
--- a/src/file-events/cpp/logging.cpp
+++ b/src/file-events/cpp/logging.cpp
@@ -18,7 +18,7 @@ Logging::~Logging() {
     env->DeleteGlobalRef(clsLogger);
 }
 
-inline bool Logging::enabled(LogLevel level) {
+bool Logging::enabled(LogLevel level) {
     return minimumLogLevel <= level;
 }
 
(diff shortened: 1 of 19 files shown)


src/file-events/headers/generic_fsnotifier.h
```diff
@@ -59,8 +60,8 @@ class AbstractServer;
 
 class Command {
 public:
-    Command() {};
-    virtual ~Command() {};
+    Command(){};
```
wolfs: I don't understand what is happening here. When I saved the file, auto-format added the spaces and now your auto-format removes them again?

src/file-events/headers/generic_fsnotifier.h
```diff
@@ -178,13 +175,13 @@ class TerminateCommand : public Command {
     }
 };
 
-struct JniConstants {
-    JniConstants(JNIEnv* env);
-    void unload(JNIEnv* env);
+class JniConstants : public JniSupport {
```
wolfs: I think we should move `JniConstants` to `jni_support` as well.

src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
```diff
@@ -624,4 +627,33 @@ class BasicFileEventFunctionsTest extends AbstractFileEventFunctionsTest {
...
+    def "can set log level"() {
+        given:
+        def nativeLogger = Logger.getLogger(NativeLogger.name)
+        def originalLevel = nativeLogger.level
+
+        when:
+        logging.clear()
+        nativeLogger.level = Level.FINEST
+        // Wait for the change to be picked up
+        Thread.sleep(1500)
```
wolfs: Can you extract the sleep to a method? So you don't need to duplicate the comment.

src/file-events/headers/generic_fsnotifier.h
lptr: Yeah, I saw this, too. Do you have a different formatter plugin maybe? I can't find which one does the thing. I have these enabled:

image
image

src/file-events/headers/generic_fsnotifier.h
lptr: I think we should rename it to `GenericClasses` or something. Or maybe move them to `AbstractServer` instead? Now it's pretty simple to store a reference to a class that you intend to use in a given context.

src/file-events/headers/generic_fsnotifier.h
wolfs: I only have those extensions: 
image

Maybe the difference is that I am using a newer version of the `C/C++` extension (`0.27.0-insiders3`?

lptr: That's one explanation. Another is that the formatter behaves differently on different OSs maybe?

src/file-events/headers/generic_fsnotifier.h
wolfs: Storing is simple, though there is only one `JNI_OnLoad` method. The method currently lives in `generic_fsnotifier.cpp`, maybe that method should go somewhere else to initialize the logging and the jni_constants?

I wouldn't move them to `AbstractServer` since I think it is better to store the classes references on load.

src/file-events/headers/generic_fsnotifier.h
lptr: Yeah, I can reproduce your formatting with the `0.27-insiders3` version. Let's keep that version.

src/file-events/headers/generic_fsnotifier.h
lptr: Alright, I moved them to `services.cpp`.

src/file-events/cpp/logging.cpp
```diff
@@ -9,6 +9,10 @@ Logging::Logging(JavaVM* jvm)
     , getLevelMethod(getThreadEnv()->GetStaticMethodID(clsLogger.get(), "getLogLevel", "()I")) {
 }
 
+void Logging::invalidateLogLevelCache() {
+    lastLevelCheck = chrono::steady_clock::time_point();
```
wolfs: What does this do? Create a new timepoint for 0ms? I didn't find documentation on the constructor of `time_point`...

lptr: > Default constructor, creates a time_point representing the Clock's epoch (i.e., `time_since_epoch()` is zero).

https://en.cppreference.com/w/cpp/chrono/time_point/time_point

PR #154, opened 2020-03-24
Add a test for hierarchical watchers deleting a hierarchy

wolfs:
This is similar to the test on Linux, only for the hierarchical
watchers.

### DIFF
--- a/src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
+++ b/src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
@@ -618,14 +618,14 @@ class BasicFileEventFunctionsTest extends AbstractFileEventFunctionsTest {
             : Platform.current().windows
             ? [event(MODIFIED, removedFile), event(REMOVED, removedFile, false), event(REMOVED, watchedDir)]
             : [event(REMOVED, removedFile), event(REMOVED, watchedDir)]
-        def directoryDeleted = removedDir.deleteDir()
+        def directoryRemoved = removedDir.deleteDir()
 
         then:
         expectedChanges.await()
-        directoryDeleted == canDeleteDirectory
+        directoryRemoved == expectDirectoryRemoved
 
         where:
-        ancestry                            | removedDirectory             | canDeleteDirectory
+        ancestry                            | removedDirectory             | expectDirectoryRemoved
         "watched directory"                 | { it }                       | true
         "parent of watched directory"       | { it.parentFile }            | !OperatingSystem.current.windows
         "grand-parent of watched directory" | { it.parentFile.parentFile } | !OperatingSystem.current.windows
--- a/src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
+++ b/src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
@@ -24,6 +24,7 @@ import spock.lang.IgnoreIf
 import spock.lang.Requires
 import spock.lang.Unroll
 import spock.util.concurrent.AsyncConditions
+import spock.util.environment.OperatingSystem
 
 import java.util.logging.Level
 import java.util.logging.Logger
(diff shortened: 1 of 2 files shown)


src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
```diff
@@ -617,16 +618,17 @@ class BasicFileEventFunctionsTest extends AbstractFileEventFunctionsTest {
             : Platform.current().windows
             ? [event(MODIFIED, removedFile), event(REMOVED, removedFile, false), event(REMOVED, watchedDir)]
             : [event(REMOVED, removedFile), event(REMOVED, watchedDir)]
-        removedDir.deleteDir()
+        def directoryDeleted = removedDir.deleteDir()
 
         then:
         expectedChanges.await()
+        directoryDeleted == canDeleteDirectory
```
lptr: Maybe call it `expectDirectoryToBeRemoved`?

src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
```diff
@@ -617,16 +618,17 @@ class BasicFileEventFunctionsTest extends AbstractFileEventFunctionsTest {
             : Platform.current().windows
             ? [event(MODIFIED, removedFile), event(REMOVED, removedFile, false), event(REMOVED, watchedDir)]
             : [event(REMOVED, removedFile), event(REMOVED, watchedDir)]
-        removedDir.deleteDir()
+        def directoryDeleted = removedDir.deleteDir()
```
lptr: Would be nice to standardize on "remove" as the term.

src/test/groovy/net/rubygrapefruit/platform/file/FileEventFunctionsStressTest.groovy
```diff
@@ -156,6 +156,43 @@ class FileEventFunctionsStressTest extends AbstractFileEventFunctionsTest {
...
+            void reportError(Throwable ex) {
+                ex.printStackTrace()
+                uncaughtFailureOnThread << ex
+            }
+        }
+
+        when:
+        def watcher = startNewWatcher(callback)
+        watcher.startWatching(watchedDir)
+        Thread.sleep(500)
```
lptr: Can we do better than sleeping here? If we want all the directories to be deleted, then we should have a latch doing that. If we want to stop in the middle of the process, we should have a latch for that. WDYT? Can we maybe do better than sleeping in other cases maybe?

lptr: Even if we leave the sleep here, there should be a comment explaining what we are waiting for. Same for other places where we sleep in stress tests.

src/test/groovy/net/rubygrapefruit/platform/file/FileEventFunctionsStressTest.groovy
```diff
@@ -156,6 +156,43 @@ class FileEventFunctionsStressTest extends AbstractFileEventFunctionsTest {
...
+        assert watchedDir.deleteDir()
+
+        then:
+        watcher.close()
```
lptr: This is a strange `then:` clause. Should this be `cleanup:` instead?

src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
wolfs: I am using `removed` now consistently.

src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
wolfs: Done.

src/test/groovy/net/rubygrapefruit/platform/file/FileEventFunctionsStressTest.groovy
wolfs: This sleep was for making sure no changes from creating the directories come in before we start deleting them. I replaced it by `waitForChangeEventLatency()`.

src/test/groovy/net/rubygrapefruit/platform/file/FileEventFunctionsStressTest.groovy
wolfs: I moved it to the `when` clause and now use `noExceptionThrown()` in the `then` clause.

wolfs:
@lptr PTAL!

PR #155, opened 2020-03-24
Allow (un)registering multiple paths at once

lptr:
On Windows and macOS typically we only want to watch a small number of paths that don't change over time, as each path is watched recursively. However, on Linux watch points need to be (un)registered frequently and en masse. The current API requires each change to be communicated to the server in a separately synchronized `Command` which is slow. Instead we want to allow the application to pass a number of watch points in at the same time.

### DIFF
--- a/src/main/java/net/rubygrapefruit/platform/file/FileWatcher.java
+++ b/src/main/java/net/rubygrapefruit/platform/file/FileWatcher.java
@@ -8,9 +8,9 @@ import java.io.IOException;
  * A handle for watching file system locations.
  */
 public interface FileWatcher extends Closeable {
-    void startWatching(File path);
+    void startWatching(File... paths);
 
-    void stopWatching(File path);
+    void stopWatching(File... paths);
 
     /**
      * Stops watching and releases any native resources.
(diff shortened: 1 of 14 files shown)


src/main/java/net/rubygrapefruit/platform/file/FileWatcher.java
```diff
@@ -8,9 +8,9 @@
  * A handle for watching file system locations.
  */
 public interface FileWatcher extends Closeable {
-    void startWatching(File path);
+    void startWatching(File... paths);
```
wolfs: Can we use a collection instead of a varargs argument here? I think that would be better.

lptr: I actually reverted from using a collection to varargs. It allows for simpler (=unmodified) code in our tests. I don't think for actual applications there is a big difference either way.

wolfs: I think there is, since we will most likely have a collection there, and then then we will need to convert the collection to an array of `File`s, then the array of `File`s will need to be converted to an array of `String`s and then passed to the native side.

lptr: Not a hill I'm willing to die on. :) Reverted to using collections.

PR #159, opened 2020-03-26
Use a BlockingQueue to pass events to the Java side

lptr:
This has a number of benefits:

1) it simplifies the code somewhat as there is no need to handle exceptions in user code on the native side,
2) ~~it simplifies test code a lot, as we don't need to handle async events,~~ (this part has been extracted into #160)
3) since the native thread is now free of user code execution, it is impossible for user code to try and run a command from this thread, preventing hard-to-debug hangs.
4) we can detect overflows to the event queue and change them into an overflow event

### DIFF
--- a/src/file-events/cpp/jni_support.cpp
+++ b/src/file-events/cpp/jni_support.cpp
@@ -33,8 +33,17 @@ JNIEnv* JniSupport::getThreadEnv() {
     return env;
 }
 
-void JniSupport::rethrowJavaException(JNIEnv* env) {
+jthrowable JniSupport::logJavaException(JNIEnv* env) {
     jthrowable exception = env->ExceptionOccurred();
+    if (exception != nullptr) {
+        env->ExceptionDescribe();
+        env->ExceptionClear();
+    }
+    return exception;
+}
+
+void JniSupport::rethrowJavaException(JNIEnv* env) {
+    jthrowable exception = logJavaException(env);
     if (exception != nullptr) {
         env->ExceptionDescribe();
         env->ExceptionClear();
--- a/src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
+++ b/src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
@@ -10,7 +10,6 @@ import java.io.InterruptedIOException;
 import java.util.Collection;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
 
@@ -78,8 +77,7 @@ public abstract class AbstractFileEventFunctions implements NativeIntegration {
         }
 
         private void queueEvent(FileWatchEvent event, boolean deliverOnOverflow) throws InterruptedException {
-            // TODO Make the timeout configurable
-            if (!eventQueue.offer(event, 1, SECONDS)) {
+            if (!eventQueue.offer(event)) {
                 NativeLogger.LOGGER.info("Event queue overflow, dropping all events");
                 signalOverflow(null);
                 if (deliverOnOverflow) {
--- a/src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
+++ b/src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
@@ -62,13 +62,33 @@ public abstract class AbstractFileEventFunctions implements NativeIntegration {
         @SuppressWarnings("unused")
         public void pathChanged(int typeIndex, String path) throws InterruptedException {
             FileWatchEvent.Type type = FileWatchEvent.Type.values()[typeIndex];
-            eventQueue.put(new ChangeEvent(type, path));
+            if (type == FileWatchEvent.Type.OVERFLOWED) {
+                signalOverflow();
+            } else {
+                queueEvent(new ChangeEvent(type, path), false);
+            }
         }
 
         // Called from the native side
         @SuppressWarnings("unused")
         public void reportError(Throwable ex) throws InterruptedException {
-            eventQueue.put(new ErrorEvent(ex));
+            queueEvent(new ErrorEvent(ex), true);
+        }
+
+        private void queueEvent(FileWatchEvent event, boolean deliverOnOverflow) throws InterruptedException {
+            // TODO Add some timeout here
+            if (!eventQueue.offer(event)) {
+                signalOverflow();
+                if (deliverOnOverflow) {
+                    eventQueue.put(event);
+                }
+            }
+        }
+
+        private void signalOverflow() throws InterruptedException {
+            NativeLogger.LOGGER.info("Event queue overflow, dropping all events");
+            eventQueue.clear();
+            eventQueue.put(new ChangeEvent(FileWatchEvent.Type.OVERFLOWED, null));
         }
     }
 
--- a/src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
+++ b/src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
@@ -12,6 +12,7 @@ import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
+import static net.rubygrapefruit.platform.file.FileWatchEvent.Type.OVERFLOWED;
 
 public abstract class AbstractFileEventFunctions implements NativeIntegration {
     public static String getVersion() {
@@ -46,8 +47,11 @@ public abstract class AbstractFileEventFunctions implements NativeIntegration {
     }
 
     /**
-     * Configures a new watcher using a builder. Call {@link AbstractWatcherBuilder#start()} to
-     * actually start the {@link FileWatcher}.
+     * Configures a new watcher using a builder.
+     * Call {@link AbstractWatcherBuilder#start()} to actually start the {@link FileWatcher}.
+     *
+     * The queue must have a total capacity of at least 2 elements.
+     * The caller should only consume events from the queue, and never add any of their own.
      */
     public abstract AbstractWatcherBuilder newWatcher(BlockingQueue<FileWatchEvent> queue);
 
@@ -61,9 +65,9 @@ public abstract class AbstractFileEventFunctions implements NativeIntegration {
 
         // Called from the native side
         @SuppressWarnings("unused")
-        public void pathChanged(int typeIndex, String path) throws InterruptedException {
+        public void pathChanged(int typeIndex, String path) {
             FileWatchEvent.Type type = FileWatchEvent.Type.values()[typeIndex];
-            if (type == FileWatchEvent.Type.OVERFLOWED) {
+            if (type == OVERFLOWED) {
                 signalOverflow(path);
             } else {
                 queueEvent(new ChangeEvent(type, path), false);
@@ -72,23 +76,41 @@ public abstract class AbstractFileEventFunctions implements NativeIntegration {
 
         // Called from the native side
         @SuppressWarnings("unused")
-        public void reportError(Throwable ex) throws InterruptedException {
+        public void reportError(Throwable ex) {
             queueEvent(new ErrorEvent(ex), true);
         }
 
-        private void queueEvent(FileWatchEvent event, boolean deliverOnOverflow) throws InterruptedException {
+        private void queueEvent(FileWatchEvent event, boolean deliverOnOverflow) {
             if (!eventQueue.offer(event)) {
                 NativeLogger.LOGGER.info("Event queue overflow, dropping all events");
                 signalOverflow(null);
                 if (deliverOnOverflow) {
-                    eventQueue.put(event);
+                    forceQueueEvent(event);
                 }
             }
         }
 
-        private void signalOverflow(@Nullable String path) throws InterruptedException {
+        private void signalOverflow(@Nullable String path) {
             eventQueue.clear();
-            eventQueue.put(new ChangeEvent(FileWatchEvent.Type.OVERFLOWED, path));
+            forceQueueEvent(new ChangeEvent(OVERFLOWED, path));
+        }
+
+        /**
+         * Queue event to a queue that we expect has enough capacity to accept the event.
+         * We expect there is enough space because we just cleared the queue, and thus
+         * it should have enough space.
+         *
+         * This can fail if the queue is extremely small (has 0 capacity, or has a capacity of
+         * 1 and we are trying to queue an error event here right after an overflow event).
+         * The queue can also be full if some other thread is adding events to it.
+         * Both a queue with a less than two element capacity and pushing events from user code
+         * are forbidden. If they occur the best we can do is log the situation.
+         */
+        private void forceQueueEvent(FileWatchEvent event) {
+            boolean eventPublished = eventQueue.offer(event);
+            if (!eventPublished) {
+                NativeLogger.LOGGER.severe("Couldn't queue event: " + event);
+            }
         }
     }
 
--- a/src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
+++ b/src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
@@ -65,7 +65,7 @@ public abstract class AbstractFileEventFunctions implements NativeIntegration {
         public void pathChanged(int typeIndex, String path) throws InterruptedException {
             FileWatchEvent.Type type = FileWatchEvent.Type.values()[typeIndex];
             if (type == FileWatchEvent.Type.OVERFLOWED) {
-                signalOverflow();
+                signalOverflow(path);
             } else {
                 queueEvent(new ChangeEvent(type, path), false);
             }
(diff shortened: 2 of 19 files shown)


test-app/src/main/java/net/rubygrapefruit/platform/test/Main.java
```diff
@@ -366,25 +368,38 @@ private static void fileSystems() {
         System.out.println("* File systems: ");
         for (FileSystemInfo fileSystem : fileSystems.getFileSystems()) {
             System.out.println(String.format("    * %s -> %s (type: %s %s, case sensitive: %s, case preserving: %s)",
-                    fileSystem.getMountPoint(), fileSystem.getDeviceName(), fileSystem.getFileSystemType(),
-                    fileSystem.isRemote() ? "remote" : "local", fileSystem.isCaseSensitive(), fileSystem.isCasePreserving()));
+                fileSystem.getMountPoint(), fileSystem.getDeviceName(), fileSystem.getFileSystemType(),
+                fileSystem.isRemote() ? "remote" : "local", fileSystem.isCaseSensitive(), fileSystem.isCasePreserving()));
```
lptr: I noticed these but I'd rather keep them. They are not very disruptive to the review, and if I revert now, we'll reformat again later.

test-app/src/main/java/net/rubygrapefruit/platform/test/Main.java
```diff
@@ -366,25 +368,38 @@ private static void fileSystems() {
...
-            @Override
-            public void reportError(Throwable ex) {
-                ex.printStackTrace();
+            public void run() {
+                while (!Thread.currentThread().isInterrupted()) {
+                    FileWatchEvent event;
+                    try {
+                        event = eventQueue.take();
+                    } catch (InterruptedException e) {
+                        break;
```
wolfs: The thread should clear the interrupted flag here, right?

lptr: That'd be prudent, yes. We could also join the thread. But it's just a daemon thread and we are about to exit the application anyway, so I didn't bother. Do you think it's important?

wolfs: It just catched my eye.

lptr:
I've updated this PR with the latest `master` using the builder pattern and Java-based threading.

It's missing a nicer type hierarchy around `FileWatchEvent`s before it can be merged. I'd like to separate failure events into their own public subtype. Let's discuss this when you are back @wolfs.

lptr:
OK, so one other advantage of this method is that we can detect an overflow of the queue in `native-platform` already, and clear it and replace the event with an overflow event.

lptr:
OK, so it seems that everything works now, but one thing: removing the watched directory on Windows seems to be running into random `OVERFLOWED` events now. If I turn logging to `FINEST` the tests pass. Otherwise some or all of the three tests in `can detect #ancestry removed` will fail with an overflow. This clearly doesn't happen on `master`.

Here's a log from a successful [`master` build](https://builds.gradle.org/buildConfiguration/GradleNative_NativePlatform_BuildWindows/33735707?buildTab=tests&pager.currentPage=1):

```log
15:27:24.291  INFO    >>> Running 'can detect watched directory removed'
15:27:24.293  INFO    > Creating ...\root\parent\removed\file.txt
15:27:24.294  INFO    < Created ...\root\parent\removed\file.txt
15:27:24.298  INFO    > Received  MODIFIED ...\root\parent\removed\file.txt
15:27:24.300  INFO    > Received  REMOVED ...\root\parent\removed
15:27:24.300  INFO    > Expecting MODIFIED ...\root\parent\removed\file.txt
15:27:24.301  INFO    > Expecting optional REMOVED ...\root\parent\removed\file.txt
15:27:24.301  INFO    > Expecting REMOVED ...\root\parent\removed
15:27:25.302  INFO    <<< Finished 'can detect watched directory removed'
```

And here it is from a [failed build on the PR](https://builds.gradle.org/buildConfiguration/GradleNative_NativePlatform_BuildWindows/33748056?mode=builds):

```log
22:59:31.365  INFO    >>> Running 'can detect parent of watched directory removed'
22:59:31.367  INFO    > Creating ...\root\parent\removed\file.txt
22:59:31.367  INFO    < Created ...\root\parent\removed\file.txt
22:59:31.368  INFO    > Expecting MODIFIED ...\root\parent\removed\file.txt
22:59:31.368  INFO    Detected overflow for C:\tcagent1\temp\buildTmp\junit6615311412972366063\can detect parent of watched directory removed\root\parent\removed
22:59:31.369  INFO    > Expecting optional REMOVED ...\root\parent\removed\file.txt
22:59:31.370  INFO    > Expecting REMOVED ...\root\parent\removed
22:59:31.371  INFO    > Received OVERFLOWED C:\tcagent1\temp\buildTmp\junit6615311412972366063\can detect parent of watched directory removed\root\parent\removed
22:59:31.371  INFO    > Received REMOVED C:\tcagent1\temp\buildTmp\junit6615311412972366063\can detect parent of watched directory removed\root\parent\removed
22:59:32.486  INFO    <<< Finished 'can detect parent of watched directory removed'

Condition failed with Exception:

expectEvents expectedEvents
|            |
|            [MODIFIED ...\root\parent\removed\file.txt, optional REMOVED ...\root\parent\removed\file.txt, REMOVED ...\root\parent\removed]
org.spockframework.util.InternalSpockError: Events received differ from expected:
 - UNEXPECTED OVERFLOWED ...\root\parent\removed
 - MISSING    MODIFIED ...\root\parent\removed\file.txt
 - MISSING    optional REMOVED ...\root\parent\removed\file.txt
 - MATCHED    REMOVED ...\root\parent\removed
	at org.spockframework.util.Assert.that(Assert.java:40)
	at net.rubygrapefruit.platform.file.AbstractFileEventFunctionsTest.expectEvents(AbstractFileEventFunctionsTest.groovy:387)
	at net.rubygrapefruit.platform.file.BasicFileEventFunctionsTest.can detect #ancestry removed(BasicFileEventFunctionsTest.groovy:685)
```

The fact that changing the log level makes a difference seems to suggest some kind of timing/locking issue.

lptr:
Note that the `REMOVED` event we get after the `OVERFLOWED` event is not received from the OS, but is sent because we explicitly report it if `listen()` fails and the directory is not present here:

https://github.com/gradle/native-platform/blob/0e6a724ee656191798fd832e6f88429cdc5d0f07/src/file-events/cpp/win_fsnotifier.cpp#L174-L180

lptr:
OK, so I can run any one of the three ancestry-deletion tests successfully, but if I try to run two or three, the subsequent tests fail. There seems to be something not properly closed or cleaned up here.

lptr:
I'm suspecting that a problem was present in the non-blocking-queue using approach, too, but it's getting triggered more often now. Perhaps that's what's causing the random overflows in other tests on Windows?

lptr:
There seems to be a strong correlation between the amount we log and the frequency of the problem happening. Maybe the difference between this PR and the `master` branch is that we log more on `master` (i.e. the starting of the thread), and that's why the problem doesn't happen there?

lptr:
OK, I think I get it now. Here in `Server::handleEvents()` we used to let the code pass further when we detected that the watched directory was removed:

https://github.com/gradle/native-platform/blob/5180a7616d7f969e344925acda1d35204427d749/src/file-events/cpp/win_fsnotifier.cpp#L135-L141

Only a few lines below this we detect that the number of transferred bytes is 0, and produce an overflow event. (And then further down we try to restart watching the removed directory _again.)_

Instead, we must return right after detecting the removal.

This makes sense and seems to give an explanation as to why we were getting overflows out of nowhere. It does not explain why the problem only occurs on this branch and not on `master`. It also doesn't explain why the `OVERFLOWED` event is reported _before_ the `REMOVED` event, and not after. Maybe that's because we explicitly clear the event queue when we report an overflow?

lptr:
Yep, that's what's happening:

1) we get a regular callback with a payload, and we report the file in the directory being `REMOVED` etc.—we start to listen again,
2) we get a callback again with the error code signaling the directory gone in `handleEvents()` (and 0 bytes transferred), so we report the watched directory to be `REMOVED` **and let the code pass through instead of returning here**
3) a few lines further down we then detect that the number of bytes transferred is 0, so we report an overflow. The reporting of the overflow clears the previous `REMOVED` event (and any others that we had reported)*
3) we then go on to `listen()` again, which detects the directory gone again, and we then report yet another `REMOVED` event.

So in the end we end up with `[OVERFLOWED, REMOVED]` in the queue, with the previous events cleared. The clearing of the previous events is a new thing on the branch, and that's why we detect this problem now, and not before, when we got the expected events.

Also, if the consuming thread is quick enough to take the events before the queue is cleared (because, say, the producing thread is slowed down by extra logging), then the test passes on this branch, too.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
```diff
@@ -46,25 +50,47 @@ public AbstractWatcherBuilder(FileWatcherCallback callback) {
...
+                queueEvent(new ChangeEvent(type, path), false);
+            }
         }
 
         // Called from the native side
         @SuppressWarnings("unused")
-        public void reportError(Throwable ex) {
-            delegate.reportError(ex);
+        public void reportError(Throwable ex) throws InterruptedException {
+            queueEvent(new ErrorEvent(ex), true);
```
wolfs, line 77: Shouldn't we clear the queue on error as well? It doesn't really make sense to process any other events before the error, right?

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
```diff
@@ -46,25 +50,47 @@ public AbstractWatcherBuilder(FileWatcherCallback callback) {
...
+        }
+
+        private void queueEvent(FileWatchEvent event, boolean deliverOnOverflow) throws InterruptedException {
+            // TODO Make the timeout configurable
+            if (!eventQueue.offer(event, 1, SECONDS)) {
```
wolfs, line 82: I am not sure if waiting here makes sense. If the blocking queue is full, then we have an overflow - let's clear it and add the overflow event. That also has the advantage that there won't be an interrupted exception.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
```diff
@@ -46,25 +50,47 @@ public AbstractWatcherBuilder(FileWatcherCallback callback) {
...
+                signalOverflow(null);
+                if (deliverOnOverflow) {
+                    eventQueue.put(event);
+                }
+            }
+        }
+
+        private void signalOverflow(@Nullable String path) throws InterruptedException {
+            eventQueue.clear();
+            eventQueue.put(new ChangeEvent(FileWatchEvent.Type.OVERFLOWED, path));
```
wolfs, line 93: I think we should use `offer` without a timeout here and fail if the queue cannot take the single event. Blocking doesn't make sense anyhow, right?

The question is what to do if the queue doesn't accept the overflow event, even when we just cleared it. Throwing an exception from here causes an error to be reported, which probably also can't be put in the queue. We could set a flag here that everything is borked and then
- try to report an error every time an event comes in
- throw an exception when closing the watcher

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
```diff
@@ -46,25 +50,47 @@ public AbstractWatcherBuilder(FileWatcherCallback callback) {
...
+            queueEvent(new ErrorEvent(ex), true);
+        }
+
+        private void queueEvent(FileWatchEvent event, boolean deliverOnOverflow) throws InterruptedException {
+            // TODO Make the timeout configurable
+            if (!eventQueue.offer(event, 1, SECONDS)) {
+                NativeLogger.LOGGER.info("Event queue overflow, dropping all events");
+                signalOverflow(null);
+                if (deliverOnOverflow) {
+                    eventQueue.put(event);
```
wolfs, line 86: Let's not block here, but fail if the queue is unable to accept the error event.

src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
```diff
@@ -514,34 +513,6 @@ class BasicFileEventFunctionsTest extends AbstractFileEventFunctionsTest {
         expectEvents change(CREATED, new File(reportedDir, fileInUppercaseDir.name))
     }
 
-    def "can handle exception in callback"() {
```
wolfs, line 517: In its current state, an interrupted exception can be thrown in the callback. We should test that this doesn't crash the JVM.

src/test/groovy/net/rubygrapefruit/platform/file/FileEventFunctionsOverflowTest.groovy
```diff
@@ -79,7 +81,45 @@ class FileEventFunctionsOverflowTest extends AbstractFileEventFunctionsTest {
...
+        waitForChangeEventLatency()
+
+        then:
+        // We still have the notification in there about the first file
+        singleElementQueue.peek().type == CREATED
+        singleElementQueue.peek().path == firstFile.absolutePath
+
+        when:
+        // Wait for the next event to time out and be replaced with an overflow event
+        Thread.sleep(1000)
```
wolfs, line 111: Can we extract this constant or the sleep into a method?

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
lptr: I'm not sure. We clear the events on overflow because with an overflow we already know we lost some events, so it's just a matter of quantity. There is no such necessity when errors arise.

If we cleared the queue on error, we'd assume that errors invalidate all previous events. While that's the approach we'll take with Gradle, it's not necessarily the only possible way to handle this.

Also, if we'd clear the queue on an error, should we also clear on an unknown event?

I think it's better as it is.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
lptr: Alright, you convinced me. I'll remove this wait.

lptr: I'll add it on top of #185.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
lptr: I like the idea to only use `offer()`. I don't like the idea of recording the invalid state and acting differently if it's set; it adds too much complexity.

We are only pushing to it from the processor thread, so it should be empty after `clear()`. The only two exceptions I have in mind is if it's a custom queue with 0 capacity, or if another thread is adding elements to the queue (which it shouldn't).

Neither of these seem like cases we need to handle in any graceful way. So throwing an `IllegalStateException` sounds good to me. WDYT?

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
lptr: Actually, I won't. Let's finish this here in this PR, and I'll rebase #185 later.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
wolfs: Throwing an `IllegalStateException` is perfectly fine with me. The problem with that is that the exception will end up in the native code which will then report the illegal state exception as an error, causing another illegal state exception since we still can't add to the queue, right?

src/test/groovy/net/rubygrapefruit/platform/file/FileEventFunctionsOverflowTest.groovy
lptr: Won't be needed if we don't wait on `offer()`.

src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
lptr: That won't happen when we don't wait of queueing the events.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
lptr: I thought about this a bit more. The best we can do in this situation I think is to log the event and then ignore the fact that we had such a problem.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
lptr: We aren't blocking anymore.

src/file-events/cpp/jni_support.cpp
```diff
@@ -33,8 +33,17 @@ JNIEnv* JniSupport::getThreadEnv() {
     return env;
 }
 
-void JniSupport::rethrowJavaException(JNIEnv* env) {
+jthrowable JniSupport::logJavaException(JNIEnv* env) {
```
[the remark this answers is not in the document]
lptr: Alright. I'll do this in the #185 PR, ok?

PR #161, opened 2020-03-27
Fix closing empty watchpoint on macOS

lptr:
The macOS native watcher emits `Path not watched:` messages upon shutdown. This is caused by incorrectly allocating a `std::vector`. This PR fixes the allocation and reverts another attempted (but not working) workaround targeted at the same problem.

### DIFF
--- a/src/file-events/cpp/apple_fsnotifier.cpp
+++ b/src/file-events/cpp/apple_fsnotifier.cpp
@@ -205,7 +205,7 @@ void Server::registerPath(const u16string& path) {
 
 void Server::unregisterPath(const u16string& path) {
     if (watchPoints.erase(path) == 0) {
-        logToJava(FINE, "Path is not watched: %s", utf16ToUtf8String(path).c_str());
+        logToJava(WARNING, "Path is not watched: %s", utf16ToUtf8String(path).c_str());
     }
 }
 
--- a/src/file-events/cpp/apple_fsnotifier.cpp
+++ b/src/file-events/cpp/apple_fsnotifier.cpp
@@ -203,10 +203,12 @@ void Server::registerPath(const u16string& path) {
         forward_as_tuple(this, threadLoop, path, latencyInMillis));
 }
 
-void Server::unregisterPath(const u16string& path) {
+bool Server::unregisterPath(const u16string& path) {
     if (watchPoints.erase(path) == 0) {
-        logToJava(WARNING, "Path is not watched: %s", utf16ToUtf8String(path).c_str());
+        logToJava(INFO, "Path is not watched: %s", utf16ToUtf8String(path).c_str());
+        return false;
     }
+    return true;
 }
 
 JNIEXPORT jobject JNICALL
--- a/src/file-events/cpp/apple_fsnotifier.cpp
+++ b/src/file-events/cpp/apple_fsnotifier.cpp
@@ -89,9 +89,11 @@ Server::Server(JNIEnv* env, jobject watcherCallback, long latencyInMillis)
 }
 
 Server::~Server() {
-    vector<u16string> paths(watchPoints.size());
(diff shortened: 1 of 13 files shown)


src/file-events/cpp/apple_fsnotifier.cpp
```diff
@@ -89,9 +89,11 @@ Server::Server(JNIEnv* env, jobject watcherCallback, long latencyInMillis)
 }
 
 Server::~Server() {
-    vector<u16string> paths(watchPoints.size());
-    for (auto& watchPoint : watchPoints) {
-        paths.push_back(watchPoint.first);
+    vector<u16string> paths;
+    paths.reserve(watchPoints.size());
+    for (auto& it : watchPoints) {
+        auto& path = it.first;
```
wolfs: Is the difference to the previous code that `path` now is a `value_type&&` instead of a `valueTyep&`? So it is copied instead of moved?

wolfs:
Can we add a test which shows what is going wrong?

src/file-events/cpp/apple_fsnotifier.cpp
lptr: No, the difference is a couple lines higher. Please see the commit comment on 8bbd66b9c51f5bba1d4b30956fcb59bf13eb4ae1. The fix is to call `vector::reserve()` separately.

wolfs: Ok, so this documentation:
> (2) fill constructor
Constructs a container with n elements. Each element is a copy of val (if provided).

actually means that the container has `n` elements, not that it has enough space for `n` elements. 🤦‍♂

lptr:
> Can we add a test which shows what is going wrong?

I made problems with closing the watcher be `WARNING`s in the log, and check for them explicitly in the tests.

src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -257,10 +257,10 @@ void Server::registerPath(const u16string& path) {
     watchRoots[watchPoint.watchDescriptor] = path;
 }
 
-void Server::unregisterPath(const u16string path) {
+void Server::unregisterPath(const u16string& path) {
     auto it = watchPoints.find(path);
     if (it == watchPoints.end() || it->second.status == FINISHED) {
-        logToJava(FINE, "Path is not watched: %s", utf16ToUtf8String(path).c_str());
+        logToJava(WARNING, "Path is not watched: %s", utf16ToUtf8String(path).c_str());
```
wolfs: I am not sure this should be a warning. That would require the client to not unregister a deleted directory to avoid the warning.

lptr: I think it makes sense that this is a `WARNING` in the context of the native file watching. The application can decide to disable it if it doesn't care about recoverable issues like this—that's pretty much what we are planning to do in `gradle/gradle`. Another app might care about these warnings, and leave the log level on `INFO`.

Another way to address this is use `INFO` for recoverable issues like this, and only use `WARNING` / `SEVERE` for non-recoverable stuff. This would be a bit strange for me, but acceptable. We should fail the tests then if `INFO` logs appear.

wolfs: We currently can't switch off logging per logger from the Gradle side. I would prefer if we switch to `INFO` logging for recoverable things.

lptr: We can do it for JUL loggers. But I'm fine with using `INFO`, too.

lptr: I'm now asserting all log messages >= `INFO`.

I'm also returning the status of `stopWatching()` and asserting that it actually stopped the watchers.

lptr:
Hm, did I just close this?

PR #166, opened 2020-04-01
Run tests with JNI checks enabled

lptr:
Add a separate test task that runs with `-Xcheck:jni` for file-events stuff. We should probably enable `pedantic`, see https://www.ibm.com/support/knowledgecenter/SSB23S_1.1.0.2019/com.ibm.java.vm.80.doc/docs/jni_debug.html

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,6 +22,8 @@ test {
         maxFailures = 10
         failOnPassedAfterRetry = true
     }
+    // See https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/clopts002.html
+    jvmArgs "-Xcheck:jni"
 }
 
 allprojects {
--- a/build.gradle
+++ b/build.gradle
@@ -16,15 +16,23 @@ versions {
 
 ext.isCiServer = System.getenv().get("CI") != null
 
-test {
+tasks.withType(Test) {
     retry {
         maxRetries = isCiServer ? 1 : 0
         maxFailures = 10
         failOnPassedAfterRetry = true
     }
-    // See https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/clopts002.html
-    jvmArgs "-Xcheck:jni"
 }
+
+def testJni = tasks.register("testJni", Test) {
+    // See https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/clopts002.html
+    it.jvmArgs "-Xcheck:jni"
+    it.useJUnit {
+        includeCategories 'net.rubygrapefruit.platform.testfixture.JniChecksEnabled'
+    }
+}
+
+check.dependsOn testJni
 
 allprojects {
     apply plugin: 'java'
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,3 @@
-import org.gradle.api.logging.StandardOutputListener
 import org.gradle.nativeplatform.platform.internal.DefaultNativePlatform
 
 plugins {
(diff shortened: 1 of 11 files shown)


build.gradle
```diff
@@ -16,14 +16,57 @@ versions {
...
+        }
+
+        @Override
+        void afterTest(TestDescriptor testDescriptor, TestResult testResult) {
+            currentTest = null
+        }
+    })
+    List<String> warningsDetected = []
+    it.logging.addStandardOutputListener({ String message ->
+        if (currentTest != null && message.startsWith("WARNING")) {
```
wolfs, line 57: So we don't detect any warnings from setup methods and the like?

src/file-events/cpp/services.cpp
```diff
@@ -2,17 +2,20 @@
 #include "generic_fsnotifier.h"
 
 JniConstants* jniConstants;
+NativeConstants* nativeContants;
```
wolfs, line 5:
```suggestion
NativeConstants* nativeConstants;
```

src/file-events/cpp/jni_support.cpp
```diff
@@ -54,3 +88,57 @@ JniThreadAttacher::~JniThreadAttacher() {
         cerr << "Failed to detach JNI from current thread: " << ret << endl;
     }
 }
+
+JniConstants::JniConstants(JavaVM* jvm)
```
wolfs, line 92: Why are we adding a separate class for this one class? Shouldn't we just have one class where we have all these constants?

src/file-events/cpp/generic_fsnotifier.cpp
```diff
@@ -282,9 +211,8 @@ JNIEXPORT void JNICALL Java_net_rubygrapefruit_platform_internal_jni_AbstractFil
     logging->invalidateLogLevelCache();
 }
 
-JniConstants::JniConstants(JavaVM* jvm)
+NativeConstants::NativeConstants(JavaVM* jvm)
```
wolfs, line 214: How are those native constants? Aren't these constants of things on the Java side?

src/file-events/cpp/services.cpp
lptr: Gah, thanks, fixed.

src/file-events/cpp/jni_support.cpp
lptr: I wanted to separate the JNI-only stuff from the native-platform related stuff.

src/file-events/cpp/generic_fsnotifier.cpp
lptr: `Native` (with capital `N`) is used to refer to the whole of `native-platform`.

build.gradle
lptr: Those methods are executed as part of the tests, so we do catch problems emitted via them.

wolfs: IIRC, `beforeTest` fires after the `setup` method. That is the reason why we didn't have the class name/method name in some of the test directories.

It should be mostly fine, though I think we won't catch any warnings from loading native platform.

lptr: I tested this and messages in `setup()` were reported.

wolfs: All good then.

lptr: I think what you mean is `beforeSpec()`, that is not included indeed.

lptr: ...but we don't use it in our tests.

src/file-events/cpp/generic_fsnotifier.cpp
wolfs: I would prefer having only `JniConstants`.

If you want to keep both classes, I think `NativePlatformJniConstants` would be a better name.

lptr: `NativePlatformJniConstants` it is now!

PR #167, opened 2020-04-01
Decrease folder depth to make sure we don't get an overflow on Windows

lptr:
It seems that overflow arrives around 4000 events.

We currently generate 2^11 = 2048 directories, and deleting each produces 1 `REMOVED` and 0-2 `MODIFIED` events (depending on Windows' taste at the moment). The resulting number of events hovers around 4000 that seems to be some kind of limit.

By halving the size of the deleted hierarchy we can avoid the overflow.

### DIFF
--- a/test-app/src/main/java/net/rubygrapefruit/platform/test/Main.java
+++ b/test-app/src/main/java/net/rubygrapefruit/platform/test/Main.java
@@ -415,7 +415,7 @@ public class Main {
                 .startWatcher(callback);
         } else if (Platform.current().isWindows()) {
             watcher = Native.get(WindowsFileEventFunctions.class)
-                .startWatcher(callback);
+                .startWatcher(64 * 1024, callback);
         } else {
             throw new RuntimeException("Only Windows and macOS are supported for file watching");
         }
--- a/src/file-events/cpp/win_fsnotifier.cpp
+++ b/src/file-events/cpp/win_fsnotifier.cpp
@@ -8,7 +8,7 @@ using namespace std;
 // WatchPoint
 //
 
-WatchPoint::WatchPoint(Server* server, const u16string& path)
+WatchPoint::WatchPoint(Server* server, size_t bufferSize, const u16string& path)
     : path(path)
     , status(NOT_LISTENING) {
     wstring pathW(path.begin(), path.end());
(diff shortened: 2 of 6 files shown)


lptr:
I don't get what's happening here.

Originally we created 2047 directories and received 4091 `REMOVED` and `MODIFIED` events before receiving an `OVERFLOW` event. See [build log](https://builds.gradle.org/viewLog.html?buildId=33143548&buildTypeId=GradleNative_NativePlatform_BuildWindows&fromExperimentalUI=true).

Now we create 1023 directories, and we receive 2043 events, but even though there are half the events (and they seem to include all the events we'll ever likely to get), we still get an `OVERFLOW` after that. See [build log](https://builds.gradle.org/viewLog.html?buildId=33170032&buildTypeId=GradleNative_NativePlatform_BuildWindows&fromExperimentalUI=true).

lptr:
It looks like in both cases we got all events followed up by an overflow. Why?

wolfs:
Maybe deleting the watched directory causes the overflow event?

lptr:
@wolfs I've increased the buffer size used in tests to 64K that seems to take care of the overflow problem. PTAL.

lptr:
@wolfs PTAL, I think 16 MB of buffer should be enough to prevent overflows altogether in stress tests. I moved back to using 16 kB for regular tests because that should be enough, and we should learn if not.

test-app/src/main/java/net/rubygrapefruit/platform/test/Main.java
```diff
@@ -415,7 +415,7 @@ private static FileWatcher createWatcher(String path, FileWatcherCallback callba
                 .startWatcher(callback);
         } else if (Platform.current().isWindows()) {
             watcher = Native.get(WindowsFileEventFunctions.class)
-                .startWatcher(callback);
+                .startWatcher(64 * 1024, callback);
```
wolfs, line 418: Why are we using 64k here and 16k in the tests?

lptr: Any of these numbers are guesswork. I used 16K for the tests as a kind of "almost too little" number, so we'd notice if anything started to fail soon for our tests that should have very few events. 4x that sounds like reasonable to avoid unnecessary overflows in the test app if there are lots of events.

PR #168, opened 2020-04-03
Add test for crash when closing the server while the event queue is blocked

wolfs:
Closing the watcher causes the destructor
of the server to throw an exception when the
terminate command runs into timeout.


lptr:
LGTM, needs ignoring before we can merge it.

PR #170, opened 2020-04-06
Move thread handling to Java

### LINKED ISSUE #162: Consider starting processor thread from Java
From [Android's JNI tips](https://developer.android.com/training/articles/perf-jni.html#threads_1):

> It's usually best to use `Thread.start()` to create any thread that needs to call in to Java code. Doing so will ensure that you have sufficient stack space, that you're in the correct `ThreadGroup`, and that you're using the same `ClassLoader` as your Java code. It's also easier to set the thread's name for debugging in Java than from native code (see `pthread_setname_np()` if you have a `pthread_t` or `thread_t`, and `std::thread::native_handle()` if you have a `std::thread` and want a `pthread_t`).

lptr:
Fixes #162.


lptr:
Superseded by #176.

PR #171, opened 2020-04-07
Continue listening after overflow

lptr:
We should not stop listening to a `WatchPoint` when an overflow event is received. Instead we should continue delivering the events after invalidating the whole watched hierarchy.

This is as it should be on Linux and macOS already, we only need to change the behavior on Windows.


src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -149,18 +149,16 @@ void Server::handleEvents(WatchPoint* watchPoint, DWORD errorCode, const vector<
             // Got a buffer overflow => current changes lost => send INVALIDATE on root
             logToJava(INFO, "Detected overflow for %s", utf16ToUtf8String(path).c_str());
             reportChange(env, FILE_EVENT_INVALIDATE, path);
```
wolfs, line 151: That explains the Windows soak test failure - since we don't receive any more events after the overflow.

lptr: Good.

PR #172, opened 2020-04-07
Test overflow handling

lptr:


src/file-events/cpp/apple_fsnotifier.cpp
```diff
@@ -145,7 +145,7 @@ void Server::handleEvents(
 }
 
 void Server::handleEvent(JNIEnv* env, char* path, FSEventStreamEventFlags flags) {
-    logToJava(FINE, "Event flags: 0x%x for %s", flags, path);
+    logToJava(FINER, "Event flags: 0x%x for '%s'", flags, path);
```
wolfs, line 148: Why are we using `FINER` here? That would be `TRACE` on the Gradle side, and we can't set this log level.

src/file-events/cpp/apple_fsnotifier.cpp
```diff
@@ -179,7 +179,6 @@ void Server::handleEvent(JNIEnv* env, char* path, FSEventStreamEventFlags flags)
         type = FILE_EVENT_UNKNOWN;
     }
 
-    logToJava(FINE, "Changed: %s %d", path, type);
```
wolfs, line 182: So all the logging now needs to happen in the callback?

src/file-events/cpp/apple_fsnotifier.cpp
lptr: That's a good point. I'll revert and instead bump the test log level to `INFO`.

src/file-events/cpp/apple_fsnotifier.cpp
lptr: We log the native event coming in just a few lines above. I wanted to kill one of them, because we had two log messages for every event.

lptr:
@wolfs PTAL.

PR #177, opened 2020-04-15
Test events in subdirectories

lptr:
Instead of only testing for events happening directly under the watched directory.


wolfs:
Does this mean you have not been able to reproduce what I saw in https://github.com/gradle/gradle/issues/12756 ?

lptr:
Yes. If a `MODIFIED` event (or any other event) would come in in 50 ms, we'd fail the test in `ensureNoMoreEvents()` called at the end of `expectEvents()`.

I even tried adding an expectation for a `MODIFIED` event, so that we'd wait the full 1 second for such an event to arrive, but it never materialized.

lptr:
NB that we _do_ get a `MODIFIED` event against the subdirectory when renaming a file: https://github.com/gradle/native-platform/pull/177/files#diff-e06f16f699cac4958768b70b17c3cdbeR304

PR #179, opened 2020-04-16
Fix and re-enable test for blocked event queue

lptr:
This test has been disabled in the past. This is a fix.


lptr:
Superseded by #184.

PR #182, opened 2020-04-16
Introduce separate OVERFLOWED event

lptr:


src/file-events/headers/generic_fsnotifier.h
```diff
@@ -23,7 +23,8 @@ enum FileWatchEventType {
     CREATED,
     REMOVED,
     MODIFIED,
-    INVALIDATE,
+    INVALIDATED,
+    OVERFLOWED,
```
wolfs, line 27: I suppose those types will go away when we move to the visitor kind of callback, right?

lptr: Yes. That's the primary motivation for the visitor-style API.

PR #185, opened 2020-04-20
Handle file system events through visitor-style handler

lptr:
This makes it easier for the client to ensure it handled all events.


src/file-events/cpp/generic_fsnotifier.cpp
```diff
@@ -121,14 +127,15 @@ bool AbstractServer::unregisterPaths(const vector<u16string>& paths) {
...
 }
 
-void AbstractServer::terminate() {
+void AbstractServer::terminate(JNIEnv* env) {
     unique_lock<mutex> terminationLock(terminationMutex);
     terminateRunLoop();
     // TODO Parametrize this
     auto status = terminated.wait_for(terminationLock, THREAD_TIMEOUT);
     if (status == cv_status::timeout) {
         throw FileWatcherException("Termination timed out");
```
wolfs, line 136: If termination times out, we don't report termination?

src/main/java/net/rubygrapefruit/platform/file/FileWatchEvent.java
```diff
@@ -4,35 +4,21 @@
...
 public interface FileWatchEvent {
 
-    /**
-     * The type of the change. When {@link Type#FAILURE}, {@link #getFailure()} returns
-     * the failure, and {@link #getPath()} returns {@code null}.
-     * Otherwise {@link #getFailure()} returns {@code null} and {@link #getPath()}
-     * returns either the path of the change, or {@code null} if a path is not known.
-     */
-    Type getType();
+    void handleEvent(Handler handler);
```
wolfs, line 7: We could use the usual `accept(Visitor)` names. I am not sure `handle` is better. I would drop the `Event` from `handleEvent` for sure.

src/main/java/net/rubygrapefruit/platform/file/FileWatchEvent.java
```diff
@@ -4,35 +4,21 @@
...
-     *
-     * <ul>
-     *     <li>whether calls to this method can be expected from a single thread or multiple different ones,</li>
-     *     <li>how actual events are reported.</li>
-     * </ul>
-     */
-    @Nullable
-    String getPath();
+    interface Handler {
+        void handleChangeEvent(ChangeType type, String absolutePath);
```
wolfs, line 10:
```suggestion
        void handleChange(ChangeType type, String absolutePath);
```

I would remove the `Event` from the `handle*` methods.

src/main/java/net/rubygrapefruit/platform/file/FileWatchEvent.java
```diff
@@ -4,35 +4,21 @@
...
 
-    /**
-     * Returns the failure encountered when {@link #getType()} is {@link Type#FAILURE}, otherwise {@code null}.
-     */
-    @Nullable
-    Throwable getFailure();
+        void handleUnknownEvent(@Nullable String absolutePath);
```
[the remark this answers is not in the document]
wolfs:
```suggestion
        void handleUnknownChange(String absolutePath);
```

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
```diff
@@ -44,6 +45,84 @@ public AbstractWatcherBuilder(BlockingQueue<FileWatchEvent> eventQueue) {
          * @see FileWatcher#startWatching(Collection)
          */
         public abstract FileWatcher start() throws InterruptedException;
+
+        /**
+         * Starts the file watcher with a thread consuming the events and
+         * feeding them to the given {@link FileWatchEvent.Handler}.
+         *
+         * The started thread will terminate once the returned
+         * {@link FileWatcher} is {@link FileWatcher#close() closed}.
```
wolfs, line 54:
```suggestion
         * {@link FileWatcher} is successfully {@link FileWatcher#close() closed}.
```

When closing doesn't work for some reason, the thread won't terminate. I think the watcher should terminate the thread on close always, even when it can't close the file watcher.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
```diff
@@ -44,6 +45,84 @@ public AbstractWatcherBuilder(BlockingQueue<FileWatchEvent> eventQueue) {
...
+         * feeding them to the given {@link FileWatchEvent.Handler}.
+         *
+         * The started thread will terminate once the returned
+         * {@link FileWatcher} is {@link FileWatcher#close() closed},
+         * or when it is {@link Thread#interrupt() interrupted}.
+         *
+         * Allows the configuration of the thread created via the given
+         * {@link ThreadConfigurator} before it is started.
+         */
+        public FileWatcher startWithHandler(ThreadConfigurator threadConfigurator, final FileWatchEvent.Handler handler) throws InterruptedException {
```
wolfs, line 76: I find it strange that the methods which start a thread still receive a blocking queue. It seems like you shouldn't care in that case.

Moreover, the method doesn't document what happens if the handler throws an exception. Currently, the thread would stop and all the remaining file events will overflow the queue. Shouldn't the watcher be stopped as well when there is an error?

All in all, it seems to me that the methods starting the thread shouldn't be in the builder, but a wrapper around the builder. Can we remove them from this PR and do those changes in a follow up if we want to support creating the consuming thread from native platform?

We should discuss how we stop the watcher and the consuming thread in an orderly fashion and how to handle the different kind of errors which can arise.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
```diff
@@ -66,33 +179,62 @@ public NativeFileWatcherCallback(BlockingQueue<FileWatchEvent> eventQueue) {
...
         public void pathChanged(int typeIndex, String path) {
-            FileWatchEvent.Type type = FileWatchEvent.Type.values()[typeIndex];
-            if (type == OVERFLOWED) {
-                signalOverflow(path);
-            } else {
-                queueEvent(new ChangeEvent(type, path), false);
+            EventType type = EventType.values()[typeIndex];
+            switch (type) {
+                case OVERFLOWED:
+                    signalOverflow(OverflowType.OPERATING_SYSTEM, path);
```
wolfs, line 185: Can we do the distinction here already on the native side, i.e. the native side calls different methods for the different `switch` branches here?

src/test/groovy/net/rubygrapefruit/platform/file/AbstractFileEventFunctionsTest.groovy
```diff
@@ -422,26 +460,57 @@ abstract class AbstractFileEventFunctionsTest extends Specification {
         }
     }
 
-    protected String shorten(FileWatchEvent event) {
-        return event.type.name() + " " + shorten(String.valueOf(event.path))
+    protected String format(FileWatchEvent event) {
+        String shortened = null
+        event.handleEvent(new FileWatchEvent.Handler() {
+            @Override
+            void handleChangeEvent(ChangeType type, String absolutePath) {
```
wolfs, line 467: One option to make this nicer is to parameterize the `Handler` with a generic variable `T`, and then have each of the `handle*` method return `T`. That way, `handleEvent` can also return `T`.

If we only need this for tests though, then the interface can also stay as is.

src/main/java/net/rubygrapefruit/platform/file/FileWatchEvent.java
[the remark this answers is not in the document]
lptr: It's not necessarily a change (for example, we could have forgotten to handle `kFSEventStreamEventFlagHistoryDone` on macOS or something), so I'd stick to `Event`.

I'll drop the `@Nullable`.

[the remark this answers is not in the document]
lptr: The reason for the `@Nullable` was to future-proof the API, but thinking about it again events should be bound to a path, other things can be reported as an error.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
```diff
@@ -44,6 +45,84 @@ public AbstractWatcherBuilder(BlockingQueue<FileWatchEvent> eventQueue) {
...
+         * or when it is {@link Thread#interrupt() interrupted}.
+         *
+         * Allows the configuration of the thread created via the given
+         * {@link ThreadConfigurator} before it is started.
+         */
+        public FileWatcher startWithHandler(ThreadConfigurator threadConfigurator, final FileWatchEvent.Handler handler) throws InterruptedException {
+            Thread processorThread = new Thread(new Runnable() {
+                @Override
+                public void run() {
+                    final AtomicBoolean terminated = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
lptr: I could, haven't thought of it.

src/file-events/cpp/generic_fsnotifier.cpp
lptr: Mhm, I guess we could report that it was incomplete.

src/main/java/net/rubygrapefruit/platform/file/FileWatchEvent.java
lptr: The reason I prefer longer, less generic names is finding usages of `accept()` in a codebase the size of Gradle take long and turn up a number of false positives.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
lptr: It should now terminate if the close is unsuccessful, i.e. it times out.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
lptr: I'm fine with moving the consuming thread to another PR, though I think you'd still want to be able to customize the queue instance used.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
```diff
@@ -44,6 +45,84 @@ public AbstractWatcherBuilder(BlockingQueue<FileWatchEvent> eventQueue) {
...
+                            public void handleTerminated() {
+                                terminated.set(true);
+                                handler.handleTerminated();
+                            }
+                        });
+                    }
+                }
+            }, "File watcher event handler");
+            threadConfigurator.configure(processorThread);
+            processorThread.start();
```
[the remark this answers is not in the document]
lptr: Reverted.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
lptr: I did that.

lptr:
Let me merge this now, please follow up if there's anything you want changed. We can take care of it in another PR.

lptr:
(Note the checks actually all passed, but GitHub's problems prevented them from being recorded correctly.)

lptr:
Merged to `master` in 42c2a8f9af78a80ba29e4da98725c4b8f57bcb90.

src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
wolfs: > I did that

You did what? There is no separate callback for the overflow and unknown events from the native side.

lptr: Looks like GitHub lost some of my changes. Fuck.

PR #186, opened 2020-04-20
Use a mutex per command

lptr:


src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -341,6 +341,7 @@ void Server::runLoop() {
 struct Command {
     Server* server;
```
wolfs, line 342: We don't need the server pointer here.

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -341,6 +341,7 @@ void Server::runLoop() {
...
     function<bool()> function;
+    mutex executionMutex;
```
wolfs, line 344: We can still share the `mutex`, let's use a `mutex*` here.

src/file-events/cpp/win_fsnotifier.cpp
lptr: Removed.

src/file-events/cpp/win_fsnotifier.cpp
lptr: I've thought about that, but this actually makes more sense to me. It's all pretty theoretical, because there will be very few calls through this anyway. But this way we don't need to hold up multiple threads trying to start/stop watching things just to schedule their command.

What's probably more important is that the `conditional_variable` is next to the `mutex` it belongs to.

wolfs: If it works, it is fine with me.

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -353,15 +353,14 @@ static void CALLBACK executeOnRunLoopCallback(_In_ ULONG_PTR info) {
     } catch (const exception&) {
         command->failure = current_exception();
     }
-    unique_lock<mutex> lock(command->server->executionMutex);
+    unique_lock<mutex> lock(command->executionMutex);
```
wolfs, line 356: Looks like I don't understand the locking in C++. Why do we need this lock at all?

lptr: It's not very intuitive I agree. This is required so that `wait_for()` always get called before `notify_all()`. Without this locking here you'd run into the reverse order, and `wait_for()` would hang.

PR #187, opened 2020-04-20
Follow up on Java threads

lptr:


src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
```diff
@@ -96,16 +99,26 @@ public void run() {
 
         @Override
         public void startWatching(Collection<File> paths) {
-            ensureOpen();
-            startWatching0(server, toAbsolutePaths(paths));
+            closeLock.readLock().lock();
```
wolfs, line 102: I am not sure we need to make the `FileWatcher` here thread safe. I think it is perfectly fine if it isn't thread safe.

lptr: I reverted that change.

PR #190, opened 2020-04-22
Reorganize reporter methods

lptr:
These are the two commits that were missing from https://github.com/gradle/native-platform/pull/185.

### DIFF
--- a/src/file-events/cpp/linux_fsnotifier.cpp
+++ b/src/file-events/cpp/linux_fsnotifier.cpp
@@ -112,7 +112,7 @@ void Server::processQueues(int timeout) {
         try {
             handleEvents();
         } catch (const exception& ex) {
-            reportError(getThreadEnv(), ex);
+            reportFailure(getThreadEnv(), ex);
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
```diff
@@ -55,40 +55,6 @@ public AbstractWatcherBuilder(BlockingQueue<FileWatchEvent> eventQueue) {
      */
     public abstract AbstractWatcherBuilder newWatcher(BlockingQueue<FileWatchEvent> queue);
 
-    enum EventType {
```
wolfs, line 58: \o/

src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -214,14 +220,12 @@ void Server::handleEvent(JNIEnv* env, const inotify_event* event) {
     } else if (IS_SET(mask, IN_MODIFY)) {
         type = MODIFIED;
     } else {
-        type = UNKNOWN;
+        logToJava(WARNING, "Unknown event 0x%x for %s", mask, utf16ToUtf8String(path).c_str());
+        reportUnknownEvent(env, path);
```
wolfs, line 224: We could report the unknown event flags to the Java side for all OSes, so we wouldn't need to dig them up from the log. Though it is probably a non issue.

PR #196, opened 2020-04-24
Ignore events for removed watch points on Linux

### LINKED ISSUE #194: Do not throw `_Map_base::at` when receiving an event for a recently deleted watchpoint
Under some circumstances, the Linux file watcher may receive a change event which is not `IN_IGNORED` for a recently removed watched point. The native side then throws an `out_of_range` exception when trying to access the path for the watch descriptor here: https://github.com/gradle/native-platform/blob/master/src/file-events/cpp/linux_fsnotifier.cpp#L184

See this [build scan](https://e.grdev.net/s/rimrty7xuumvm/tests/ykfkc5kxrsfs4-u6vxcjrarml7u). `SourceDependencyCleanupIntegrationTest` seems to expose the problem on the Gradle side.

wolfs:
Events different to `IN_IGNORED` can come in for
recently removed watch points.

We should handle those events gracefully.

Fixes #194

### DIFF
--- a/src/file-events/cpp/linux_fsnotifier.cpp
+++ b/src/file-events/cpp/linux_fsnotifier.cpp
@@ -181,7 +181,16 @@ void Server::handleEvent(JNIEnv* env, const inotify_event* event) {
         return;
     }
 
-    auto path = watchRoots.at(event->wd);
+    auto it = watchRoots.find(event->wd);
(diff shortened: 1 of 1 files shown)


lptr:
Superseded by https://github.com/gradle/native-platform/commit/e87da4d6c35155601c3299255ae462de24b94093

PR #198, opened 2020-04-27
Use an enum class to represent ChangeType

lptr:
This is a C++11 feature that does not pollute the top-level scope.

### DIFF
--- a/src/file-events/cpp/logging.cpp
+++ b/src/file-events/cpp/logging.cpp
@@ -22,7 +22,7 @@ bool Logging::enabled(LogLevel level) {
         rethrowJavaException(env);
         lastLevelCheck = current;
     }
-    return minimumLogLevel <= level;
+    return minimumLogLevel <= static_cast<int>(level);
 }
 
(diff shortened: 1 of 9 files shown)


src/file-events/cpp/logging.cpp
```diff
@@ -22,7 +22,7 @@ bool Logging::enabled(LogLevel level) {
         rethrowJavaException(env);
         lastLevelCheck = current;
     }
-    return minimumLogLevel <= level;
+    return minimumLogLevel <= static_cast<int>(level);
```
wolfs, line 25:
```suggestion
    return minimumLogLevel <= reinterpret_cast <int>(level);
```

Do we actually need to cast here?

lptr: We do need to cast, because `enum class` is not an `int`. The right cast to use is `static_cast`. `reinterpret_cast` is weaker, and doesn't work here: https://stackoverflow.com/a/12228616/104894

Issue #195, opened 2020-04-24
Ignore `EINVAL` when unregistering a watch point on Linux


wolfs:
When a watched directory is deleted on Linux, the watcher eventually receives the `IN_IGNORED`. It is possibly though, that the client tries to stop watching the deleted directory before the `IN_IGNORED` event is processed on the native side. This causes an exception to be thrown, since `inotify_rm_watch` fails with `EINVAL`.

For example, `org.gradle.vcs.git.internal.SourceDependencyCleanupIntegrationTest` is pretty flaky because of this: [test trend](https://e.grdev.net/scans/tests?list.size=50&list.sortColumn=startTime&list.sortOrder=desc&search.buildToolType=gradle&search.buildToolType=maven&search.relativeStartTime=P7D&search.tags=master&search.tasks=vfsRetentionTest&tests.container=org.gradle.vcs.git.internal.SourceDependencyCleanupIntegrationTest&tests.sortField=FAILED&tests.unstableOnly=true)

You can see the problem in [this build scan](https://e.grdev.net/s/y5uox237qbxbg/tests/ykfkc5kxrsfs4-u6vxcjrarml7u) with extra logging:
```
Trying to find path for 5, 0x200
Path found <project-dir>/subprojects/version-control/build/tmp/test files/SourceDependencyCleanupIntegrationTest/7pwqu/.gradle/vcs-1/dep_2p2oqdno0w5lcbejcfecouctm/dep
Watch point for path found <project-dir>/subprojects/version-control/build/tmp/test files/SourceDependencyCleanupIntegrationTest/7pwqu/.gradle/vcs-1/dep_2p2oqdno0w5lcbejcfecouctm/dep

...

Trying to find path for 6, 0x8000
Path found <project-dir>/subprojects/version-control/build/tmp/test files/SourceDependencyCleanupIntegrationTest/7pwqu/.gradle/vcs-1/dep_2p2oqdno0w5lcbejcfecouctm/dep/build/tmp/jar
Watch point for path found <project-dir>/subprojects/version-control/build/tmp/test files/SourceDependencyCleanupIntegrationTest/7pwqu/.gradle/vcs-1/dep_2p2oqdno0w5lcbejcfecouctm/dep/build/tmp/jar
[...]

Issue #197, opened 2020-04-27
Test stopping watchers on Linux while directory hierarchy is being deleted


lptr:
Currently we test closing the watcher while the directory is being deleted, which is not the same, as it disables receiving events altogether.

We should probably do the closing and the deletion from two separate threads in the existing test, too.

lptr:
We have this now.

PR #199, opened 2020-04-27
Do not fail if cancelling watch point returns EINVAL

lptr:
This is a sign that the watch point has been removed recently and we should treat it as a normal event and not fail.

Fixes https://github.com/gradle/native-platform/issues/195.

### DIFF
--- a/src/file-events/cpp/linux_fsnotifier.cpp
+++ b/src/file-events/cpp/linux_fsnotifier.cpp
@@ -19,15 +19,22 @@ WatchPoint::WatchPoint(const u16string& path, shared_ptr<Inotify> inotify, int w
     , path(path) {
 }
 
-bool WatchPoint::cancel() {
+CancelResult WatchPoint::cancel() {
     if (status == WatchPointStatus::CANCELLED) {
-        return false;
+        return CancelResult::ALREADY_CANCELLED;
     }
     status = WatchPointStatus::CANCELLED;
     if (inotify_rm_watch(inotify->fd, watchDescriptor) != 0) {
-        throw FileWatcherException("Couldn't stop watching", path, errno);
+        switch (errno) {
+            case EINVAL:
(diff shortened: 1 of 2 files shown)


src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -19,15 +19,22 @@ WatchPoint::WatchPoint(const u16string& path, shared_ptr<Inotify> inotify, int w
...
+CancelResult WatchPoint::cancel() {
     if (status == WatchPointStatus::CANCELLED) {
-        return false;
+        return CancelResult::ALREADY_CANCELLED;
     }
     status = WatchPointStatus::CANCELLED;
     if (inotify_rm_watch(inotify->fd, watchDescriptor) != 0) {
-        throw FileWatcherException("Couldn't stop watching", path, errno);
+        switch (errno) {
+            case EINVAL:
```
wolfs, line 29: Did you manage to add a test for this problem?

src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -19,15 +19,22 @@ WatchPoint::WatchPoint(const u16string& path, shared_ptr<Inotify> inotify, int w
...
+                logToJava(LogLevel::INFO, "Couldn't stop watching %s (probably because the directory was removed)", utf16ToUtf8String(path).c_str());
```
wolfs, line 30: I wonder if the INFO log level is too high...

src/file-events/cpp/linux_fsnotifier.cpp
lptr: Not yet. I wanted to get the fix in, and then add a test for closing watchers while directories are being deleted in #197.

src/file-events/cpp/linux_fsnotifier.cpp
lptr: We do `INFO` for other unexpected events, too (like `Path is not watched`).

PR #201, opened 2020-05-01
Enable symlink tests for Windows

lptr:


src/test/groovy/net/rubygrapefruit/platform/file/SymlinkFileEventFunctionsTest.groovy
```diff
@@ -96,4 +164,10 @@ class SymlinkFileEventFunctionsTest extends AbstractFileEventFunctionsTest {
         ex.message == "Already watching path: ${canonicalDir.absolutePath}"
     }
 
+    private void createSymlink(File linked, File target) {
+        LOGGER.info("> Creating link to ${shorten(target)} in ${shorten(linked)}")
```
wolfs, line 168: Does the logging help much? If not, can we remove this method?

lptr: I have a similar method for `createNewFile()` that did help a lot.

PR #202, opened 2020-05-04
Improve Linux and macOS stat performance

wolfs:
We should be able to be as fast as Java NIO when stating files.

### DIFF
--- a/src/main/cpp/posix.cpp
+++ b/src/main/cpp/posix.cpp
@@ -36,6 +36,8 @@
 #include <termios.h>
 #include <unistd.h>
 
+jmethodID fileStatDetailsMethodId;
+
 JNIEXPORT void JNICALL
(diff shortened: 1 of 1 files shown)


wolfs:
See https://github.com/gradle/native-platform/issues/203 for the situation on Windows file systems.

wolfs:
The following benchmarks were done with the jmh benchmark in the `:native` subproject of `gradle/gradle`:

macOS before:
image

macOS after:
image

Especially the missing file performance increases significantly.

wolfs:
> LGTM as a quickfix. Let's share `JniSupport` for these later, though. I'm sure there are other places where it could come in handy.

Yeah, the file I changed already contains some other method lookups, it would be great to cache them all on load.

wolfs:
For some reason, Linux still seems to lag behind the Java NIO implementation and the missing file check with `File.exists()`:

Linux after:

image

Linux before:

image

PR #211, opened 2020-05-12
Handle all log levels

lptr:
For some reason we didn't handle ALL and OFF. Now we do.


src/main/java/net/rubygrapefruit/platform/internal/jni/NativeLogger.java
```diff
@@ -9,13 +9,15 @@
     static final Logger LOGGER = Logger.getLogger(NativeLogger.class.getName());
 
     enum LogLevel {
+        ALL(Level.ALL),
```
wolfs, line 12: I think we need to update the `LogLevel` enum on the native side as well, right?

lptr: You're right. Should be fixed.

PR #212, opened 2020-05-13
Fix invalid watch descriptor on Linux

lptr:


lptr:
I did click on "Merge" the first time, but for some reason we ended up closing the PR without merging it. WTF?!

PR #219, opened 2020-05-19
Archive PDB files for published artifacts

wolfs:


wolfs:
You can see the archived `pdb` files here: https://builds.gradle.org/buildConfiguration/GradleNative_NativePlatform_Publishing_PublishWindowsSnapshot/35044060?buildTab=artifacts

PR #223, opened 2020-06-12
Handle exception thrown by CloseHandle Windows API

lptr:
[`CloseHandle()` can throw](https://docs.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-closehandle#return-value) a [structured exception (SEH)](https://docs.microsoft.com/en-us/cpp/cpp/structured-exception-handling-c-cpp?view=vs-2019) which we must handle in `WatchPoint::close()` instead of letting it bubble up to `~WatchPoint()`.


lptr:
OK, so this doesn't seem to solve the problem, but I think we should merge it anyway, as the PR makes the code better.

PR #225, opened 2020-06-15
Fix Windows crash, take 3

lptr:


src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -61,9 +61,8 @@ bool WatchPoint::cancel() {
 
 WatchPoint::~WatchPoint() {
     try {
-        if (cancel()) {
-            SleepEx(0, true);
-        }
+        cancel();
```
wolfs, line 64: I wonder if the should try cancelling again here when an earlier cancel failed, i.e. not always  set the status to `FINISHED` in `cancel`. WDYT?

lptr: We could also do that by not observing the status in `cancel()`... That said, let's see if this fixes the errors, and find an explanation of why it does if it does first.

src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -322,32 +321,19 @@ void Server::runLoop() {
...
+            }
         }
     }
 
-    // If there are any pending watchers, wait for them to finish
-    if (pendingWatchPoints > 0) {
-        logToJava(LogLevel::FINE, "Waiting for %d pending watch points to finish", pendingWatchPoints);
-        SleepEx(0, true);
-    }
+    logToJava(LogLevel::FINE, "Waiting for any pending watch points to abort completely", NULL);
```
wolfs, line 335: Shall we remove this log? Do you think it adds some extra value?

lptr: I think you're right. Let's remove it separately, though.

PR #228, opened 2020-06-19
Upgradle to 6.5

wolfs:


wolfs:
@lptr I enabled file-system watching. Let's see if anything breaks. I also converted `setting.gradle` to Kotlin, so we don't need to manage the Gradle Enterprise version. Please take another look.

PR #231, opened 2020-06-23
Throw specialized exceptions for Linux inotify limits

wolfs:
We now throw specialized exception when we reach the inotify limits on Linux. This can then be handled by clients of native platform to display helpful messages.

### DIFF
--- a/src/file-events/cpp/generic_fsnotifier.cpp
+++ b/src/file-events/cpp/generic_fsnotifier.cpp
@@ -44,16 +44,12 @@ FileWatcherException::FileWatcherException(const string& message)
     : runtime_error(message) {
 }
 
+JavaExceptionThrownException::JavaExceptionThrownException()
+    : runtime_error("Java exception thrown from native code") {
+}
+
 InsufficientResourcesFileWatcherException::InsufficientResourcesFileWatcherException(const string& message)
     : FileWatcherException(message) {
-}
-
-InotifyInstanceLimitTooLowException::InotifyInstanceLimitTooLowException()
-    : InsufficientResourcesFileWatcherException("Inotify instance limit too low") {
-}
-
-InotifyWatchesLimitTooLowException::InotifyWatchesLimitTooLowException()
-    : InsufficientResourcesFileWatcherException("Inotify watches limit too low") {
 }
 
 AbstractServer::AbstractServer(JNIEnv* env, jobject watcherCallback)
@@ -198,8 +194,8 @@ Java_net_rubygrapefruit_platform_internal_jni_AbstractFileEventFunctions_00024Na
         vector<u16string> paths;
         javaToUtf16StringArray(env, javaPaths, paths);
         server->registerPaths(paths);
-    } catch (const InotifyWatchesLimitTooLowException& e) {
-        rethrowAsJavaException(env, e, nativePlatformJniConstants->inotifyWatchesLimitTooLowExceptionClass.get());
+    } catch (const JavaExceptionThrownException& e) {
+        // Ignore, the Java exception has already been thrown.
     } catch (const exception& e) {
         rethrowAsJavaException(env, e);
     }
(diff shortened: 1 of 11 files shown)


wolfs:
I tested the exception manually:

- [InotifyWatchesLimitTooLowException](https://e.grdev.net/s/c5nymixdmmnqm/tests/:test/net.rubygrapefruit.platform.file.FileEventFunctionsStressTest/can%20stop%20watching%20many%20directories%20while%20they%20are%20being%20deleted?focused-exception-line=0-0#1)
- [InotifyInstanceLimitTooLowException](https://e.grdev.net/s/5rc5mjcgs2vwq/tests/:test/net.rubygrapefruit.platform.file.BasicFileEventFunctionsTest/can%20set%20log%20level%20by%20waiting%20for%20log%20level%20cache%20to%20time%20out?focused-exception-line=0-0#1)

src/file-events/cpp/generic_fsnotifier.cpp
```diff
@@ -44,6 +44,14 @@ FileWatcherException::FileWatcherException(const string& message)
     : runtime_error(message) {
 }
 
+JavaExceptionThrownException::JavaExceptionThrownException()
+    : runtime_error("Java exception thrown from native code") {
+}
```
lptr, line 49: Let's add some comments/docs here explaining what this is.

src/file-events/cpp/generic_fsnotifier.cpp
```diff
@@ -236,5 +250,7 @@ Java_net_rubygrapefruit_platform_internal_jni_AbstractFileEventFunctions_invalid
 
 NativePlatformJniConstants::NativePlatformJniConstants(JavaVM* jvm)
     : JniSupport(jvm)
-    , nativeExceptionClass(getThreadEnv(), "net/rubygrapefruit/platform/NativeException") {
+    , nativeExceptionClass(getThreadEnv(), "net/rubygrapefruit/platform/NativeException")
+    , inotifyWatchesLimitTooLowExceptionClass(getThreadEnv(), "net/rubygrapefruit/platform/internal/jni/InotifyWatchesLimitTooLowException")
+    , inotifyInstanceLimitTooLowExceptionClass(getThreadEnv(), "net/rubygrapefruit/platform/internal/jni/InotifyInstanceLimitTooLowException") {
```
lptr, line 255: These should be moved to a separate `linuxJniConstants` object.

src/main/java/net/rubygrapefruit/platform/file/FileWatcher.java
```diff
@@ -11,7 +13,7 @@
  */
 @NotThreadSafe
 public interface FileWatcher {
-    void startWatching(Collection<File> paths);
+    void startWatching(Collection<File> paths) throws InsufficientResourcesForWatchingException;
```
lptr, line 16: Why mention a runtime exception here? And why this specifically, when we can also throw `NativeException` in general, too.

src/file-events/cpp/generic_fsnotifier.cpp
wolfs: I moved the constants. I am not sure it is worth it. Please take a look at the commit and tell me what you think.

lptr: It's ugly I agree, but I still think it's better this way. If you feel strongly, I'm okay with reverting, too.

src/main/java/net/rubygrapefruit/platform/file/FileWatcher.java
wolfs: The mention is for documentation purposes only. I added the `InsufficientResourcesForWatchingException` since that is something you may want to handle separately for this method.

src/file-events/cpp/generic_fsnotifier.cpp
wolfs: I added some documentation in the header file.

wolfs:
@lptr PTAL!

src/file-events/cpp/generic_fsnotifier.cpp
wolfs: Let's leave it as it is now.

PR #238, opened 2020-07-06
Use a single FSEventStream

lptr:
Previously we started an event stream for each individual directory hierarchy we watched. This seems to be overly expensive and there's also a limit of ~1000 event streams per machine that we might hit. So instead we are now using a single event stream that watches multiple directory hierarchies.

This makes adding/removing watches slightly slower.

There's also a change in behavior wrt watching the same directory both directly and via a symlink: in this case changes are not reported twice with the same path anymore, which is arguably better.

### DIFF
--- a/src/file-events/cpp/apple_fsnotifier.cpp
+++ b/src/file-events/cpp/apple_fsnotifier.cpp
@@ -197,6 +197,7 @@ void Server::handleEvents(
     try {
         for (size_t i = 0; i < numEvents; i++) {
             const FSEventStreamEventFlags flags = eventFlags[i];
+            const FSEventStreamEventId eventId = eventIds[i];
             if (IS_SET(flags, kFSEventStreamEventFlagHistoryDone)) {
                 // Mark all new watch points as able to receive historical events from this point on
                 for (auto& it : watchPoints) {
@@ -205,12 +206,12 @@ void Server::handleEvents(
                     }
                 }
                 finishedProcessingHistoricalEvents = true;
+                logToJava(LogLevel::FINE, "Finished processing historical events (ID %d)", eventId);
                 continue;
             }
 
             FSEventStreamEventFlags normalizedFlags = flags & ~IGNORED_FLAGS;
             const char* path = eventPaths[i];
-            const FSEventStreamEventId eventId = eventIds[i];
             logToJava(LogLevel::FINE, "Event flags: 0x%x (normalized: 0x%x) with ID %d for '%s'",
                 flags, normalizedFlags, eventId, path);
 
--- a/src/file-events/cpp/apple_fsnotifier.cpp
+++ b/src/file-events/cpp/apple_fsnotifier.cpp
@@ -26,12 +26,9 @@ void Server::createEventStream() {
         // CFRelease(cfPath);
     }
 
-    // Make sure we don't miss any events, even if we restart watching
-    // before the first events are processed
-    if (lastSeenEventId == kFSEventStreamEventIdSinceNow) {
-        lastSeenEventId = FSEventsGetCurrentEventId();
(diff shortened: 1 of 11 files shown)


src/file-events/cpp/apple_fsnotifier.cpp
```diff
@@ -160,75 +176,146 @@ static const FSEventStreamEventFlags IGNORED_FLAGS = kFSEventStreamCreateFlagNon
...
                    kFSEventStreamEventFlagItemInodeMetaMod    // file locked
                        | kFSEventStreamEventFlagItemFinderInfoMod
                        | kFSEventStreamEventFlagItemChangeOwner
                        | kFSEventStreamEventFlagItemXattrMod)) {
         type = ChangeType::MODIFIED;
-    } else if (IS_SET(normalizedFlags, kFSEventStreamEventFlagItemCreated)) {
+    } else if (IS_SET(flags, kFSEventStreamEventFlagItemCreated)) {
         type = ChangeType::CREATED;
     } else {
-        logToJava(LogLevel::WARNING, "Unknown event 0x%x (normalized: 0x%x) for %s", flags, normalizedFlags, path);
```
wolfs, line 204: Why did you remove the WARNING logging for unknown events?

src/file-events/cpp/apple_fsnotifier.cpp
```diff
@@ -4,70 +4,102 @@
...
-    // According to this comment we should not use flush at all, and it's probably broken:
-    // https://github.com/nodejs/node/issues/854#issuecomment-294892950
-    // As the comment mentions, even Watchman doesn't flush:
-    // https://github.com/facebook/watchman/blob/b397e00cf566f361282a456122eef4e909f26182/watcher/fsevents.cpp#L276-L285
-    // FSEventStreamFlushSync(watcherStream);
-    FSEventStreamStop(watcherStream);
-    FSEventStreamInvalidate(watcherStream);
-    FSEventStreamRelease(watcherStream);
+
+    FSEventStreamFlushSync(eventStream);
```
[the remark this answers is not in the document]
lptr: Yes, it somehow seems to work now. I definitely didn't get the printouts mentioned in the removed comments.

One explanation is that I used to use `kFSEventStreamEventIdSinceNow` to start the stream, which is `-1`, so my theory is that if no events arrived, the `last_id` stayed `-1` and thus we got the printout. We now always start streams with a specific event ID, so it's never less than `0`.

src/file-events/cpp/apple_fsnotifier.cpp
lptr: This was a redundant log we did on macOS, in order to capture the flags on `WARNING` level in case we encounter an unknown event. I don't think it's necessary to do it anymore, since we haven't seen any unknown events for a long time. If we do, we can use the `FINE` log above to debug it.

src/test/groovy/net/rubygrapefruit/platform/file/FileEventFunctionsStressTest.groovy
```diff
@@ -192,7 +192,7 @@ class FileEventFunctionsStressTest extends AbstractFileEventFunctionsTest {
         def onslaught = new OnslaughtExecuter(
             (otherDirs.collect { otherDir ->
                 { ->
-                    Thread.sleep((long) (Math.random() * 100))
+                    Thread.sleep((long) (Math.random() * 100) + 100)
```
lptr, line 195: This is to avoid the situation where we'd restart watching multiple times even before changes started trickling in, which is not what we want to watch here.

lptr:
Let me merge this and we can do any followup on those last couple commits in a new PR if necessary.

PR #240, opened 2020-07-09
Run compatibility tests on CentOs

wolfs:
This PR configures a build on CentOS, so it uses the already built artifacts from the upstream builds to run the native platform test on CentOs 8. It replaces the artifacts built in the current build by the artifacts from an external repository via dependency substitution, so the native libraries are not rebuilt on CentOS.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,7 @@ versions {
 }
 
 ext.isCiServer = System.getenv().get("CI") != null
+ext.testVersionFromLocalRepository = project.hasProperty('testVersionFromLocalRepository')
 
 tasks.withType(Test) {
     systemProperty "test.directory", layout.buildDirectory.dir("test files").map { it.asFile.absolutePath }.get()
@@ -23,6 +24,8 @@ tasks.withType(Test) {
         maxFailures = 10
         failOnPassedAfterRetry = true
     }
+    classpath = files(configurations.testRuntimeClasspath)
+    classpath.from(sourceSets.test.output)
 }
 
 def testJni = tasks.register("testJni", Test) {
@@ -86,7 +89,23 @@ allprojects {
 
 dependencies {
     compileOnly 'com.google.code.findbugs:jsr305:3.0.2'
+    testImplementation project(":")
     testImplementation 'org.spockframework:spock-core:1.3-groovy-2.5'
+}
+
+if (testVersionFromLocalRepository) {
+    group = "new-group-for-root-project"
+    configurations.all {
+        resolutionStrategy.dependencySubstitution {
+            substitute project(':') with module("net.rubygrapefruit:native-platform:${version}")
+        }
+    }
+    repositories {
+        maven {
+            name = "IncomingLocalRepository"
+            url = rootProject.file("incoming-repo")
+        }
+    }
 }
 
 // Only depend on variants which can be built on the current machine
@@ -443,8 +462,10 @@ afterEvaluate {
             binary.tasks.withType(LinkSharedLibrary) { builderTask ->
                 nativeJar.into("net/rubygrapefruit/platform/$variantName") { from builderTask.linkedFile }
             }
-            project.tasks.test {
-                classpath.from nativeJar
+            if (!testVersionFromLocalRepository) {
+                project.tasks.test {
+                    classpath.from nativeJar
+                }
             }
         }
     }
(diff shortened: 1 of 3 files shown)


wolfs:
This build scan shows that the compatibility test uses the libraries from the binary repository: https://ge.gradle.org/s/popr6lh2wgk7w/dependencies?focusedDependency=WzAsNywwLFswLDcsWzBdXV0&toggled=W1swXSxbMCw3XSxbMCw3LFswXV1d

build.gradle
```diff
@@ -15,6 +15,7 @@ versions {
 }
 
 ext.isCiServer = System.getenv().get("CI") != null
+ext.testVersionFromLocalRepository = project.hasProperty('testVersionFromLocalRepository')
```
lptr, line 18: Should we use some configuration-cache compatible way to get this property?

build.gradle
```diff
@@ -23,6 +24,8 @@ tasks.withType(Test) {
         maxFailures = 10
         failOnPassedAfterRetry = true
     }
+    classpath = files(configurations.testRuntimeClasspath)
+    classpath.from(sourceSets.test.output)
```
lptr, line 28: I think this needs some explanation. Can you add some comments describing why we need this?
Also, why not add the two entries the same way? Say:

```groovy
classpath = files(configurations.testRuntimeClasspath, sourceSets.test.output)
```

...or perhaps:

```groovy
classpath = files()
    .from(configurations.testRuntimeClasspath)
    .from(sourceSets.test.output)
```

build.gradle
```diff
@@ -86,9 +89,25 @@ allprojects {
 
 dependencies {
     compileOnly 'com.google.code.findbugs:jsr305:3.0.2'
+    testImplementation project(":")
```
lptr, line 92: Why is this needed?

build.gradle
lptr: Also, let's add some comment here describing the purpose of this parameter.

build.gradle
```diff
@@ -86,9 +89,25 @@ allprojects {
...
     testImplementation 'org.spockframework:spock-core:1.3-groovy-2.5'
 }
 
+if (testVersionFromLocalRepository) {
+    group = "new-group-for-root-project"
```
lptr, line 97: This looks like a considerable amount of hack that should be documented. Why do we need this?

build.gradle
```diff
@@ -443,8 +462,10 @@ afterEvaluate {
             binary.tasks.withType(LinkSharedLibrary) { builderTask ->
                 nativeJar.into("net/rubygrapefruit/platform/$variantName") { from builderTask.linkedFile }
             }
-            project.tasks.test {
-                classpath.from nativeJar
+            if (!testVersionFromLocalRepository) {
+                project.tasks.test {
+                    classpath.from nativeJar
+                }
             }
```
lptr, line 469: Can we turn this into lazy configuration somehow?

build.gradle
wolfs: We change the wiring of the test task for all runs:
Normally, the test task uses:
- output of the `main` source set
- output of the `test` source set
- testRuntimeClasspath

It is impossible to use dependency substitution for replacing the output of the main source set. So we remove the output of the main source set from the test classpath and replace it by a dependency on `project(':')` instead. That is why we need to add the dependency on project(':') here.

build.gradle
wolfs: It seems like you cannot substitute a project dependency for an external dependency with the same GAV coordinates (known limitation of the engine). So I change the GAV coordinates of the root project so the dependency substitution takes effect.

build.gradle
wolfs: Ah, so you want to avoid realizing the test task? I think I can do that.

build.gradle
wolfs: Both things done.

build.gradle
wolfs: Done.

build.gradle
wolfs: I added some comments explaining what we do there.

build.gradle
wolfs: I added a comment.

build.gradle
wolfs: Done.

wolfs:
@lptr PTAL!

PR #242, opened 2020-07-14
Remove unnecessary logging of normalized flags on macOS

lptr:


src/file-events/cpp/apple_fsnotifier.cpp
```diff
@@ -223,17 +221,17 @@ void Server::handleEvents(
             if (!finishedProcessingHistoricalEvents) {
                 WatchPointState state = getWatchPointState(pathStr);
                 if (state == WatchPointState::NEW) {
-                    logToJava(LogLevel::FINE, "Ignoring historical event %d for '%s'", eventId, path);
+                    logToJava(LogLevel::FINE, "Ignoring historical event (ID %d)", eventId);
```
wolfs, line 224: Why don't we log the path here any more?

lptr: The documentation explicitly says that the path should be ignored for this event, so logging it makes little sense.

PR #243, opened 2020-07-15
Close event stream before modifying watch point registrations

lptr:
Previously we updated event streams after the changes were made. This could cause race conditions when changes keep coming in for the unregistered events.


lptr:
I was thinking about why the problem didn't occur in our tests in `native-platform` itself. I think it's because while we have tests that check if we can stop watching while changes are being made to the watched data, it's mostly about the watched directories themselves being deleted. These are reported as different events that don't require the loopup. Maybe.

PR #245, opened 2020-07-16
Add test for watching many directories at once

lptr:


wolfs:
The only problem is that the tests are failing: on the Linux platforms we don't seem to have enough inotify watches, and there is a JVM crash on Windows: https://builds.gradle.org/buildConfiguration/GradleNative_NativePlatform_BuildWindowsAmd64/36700253?buildTab=overview

PR #246, opened 2020-07-16
Automatically publish on Bintray

wolfs:
So there is no manual intervention necessary to make artifacts visible to the public when publishing to Bintray.


wolfs:
Here is a build scan of a build discarding an uploaded milestone on Bintray: https://ge.gradle.org/s/wgnosjbbo5ax6

PR #249, opened 2020-07-21
Fix watch point mutation locking

lptr:
We've been using `mutex` in a bad way:

* on Windows we don't even need to guard `watchPoints` as all the mutation happens on the run loop thread anyway, so that's removed now,
* for macOS and Linux we should have been using `recursive_mutex` instead, as it allows reentry—this is also fixed now.

After #238 has been reverted in #248, we can still salvage a number of improvements that happened as part of the original PR. These are ported here, too.

### DIFF
--- a/src/file-events/cpp/apple_fsnotifier.cpp
+++ b/src/file-events/cpp/apple_fsnotifier.cpp
@@ -95,7 +95,7 @@ void Server::initializeRunLoop() {
 void Server::runLoop() {
     CFRunLoopRun();
 
-    unique_lock<mutex> lock(mutationMutex);
+    unique_lock<recursive_mutex> lock(mutationMutex);
     watchPoints.clear();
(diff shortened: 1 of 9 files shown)


lptr:
@wolfs: this is what we should release as `0.22-milestone-6` as it fixes the lock situation. There is now no locking around watch point mutation on Windows anymore. On Linux and macOS we use a `recursive_mutex` to guard the watch points. I wonder if the latter will remove some problems on its own.

lptr:
Note that the commands on Windows and termination all use regular `mutex`s as we need those to pass data between the threads via `condition_variable`s. This means that none of the public methods should ever be called from the processing thread, which is how things work now, so no problem.

lptr:
Yes, locking the same `mutex` twice on Windows reliably produces the `device or resource busy` exception.

PR #255, opened 2020-08-06
Move native platform code from root project to subproject

wolfs:
We move all the code from the root project to the `native-platform` subproject to make it easier to split up `native-platform`.

The build logic from the `allprojects` block in the root `build.gradle` has been moved to a plugin in `buildSrc`. Versioning now uses Gradle properties from the command line or in `gradle.properties`, instead of configuration in the root build script.


wolfs:
@lptr PTAL!

wolfs:
I added a description to the PR. I think the next follow up on this would be to move the setup for the native code in `native-platform/build.gradle` to a plugin as well, so we could use the same setup in a different subproject.

settings.gradle.kts
```diff
@@ -5,8 +5,7 @@ plugins {
 
 rootProject.name = "native-platform"
 
-include("testApp")
-
-project(":testApp").projectDir = File(rootDir, "test-app")
+include("test-app")
+include("native-platform")
```
lptr, line 9: Does this mean that we now have two projects called `native-platform`? Is that a problem?

wolfs: The root project name seems to be a different beast as the other project names. The path for the root project is still `:`, so I don't think this is a problem.

@jjohannes do you know if this could cause problems?

jjohannes: I am not aware of any problems.

We only sometimes have issues if two projects have the same "simple name" if they do stuff based on that name only (like naming a jar file, or putting it into the GAV coordinate).

But since the root project intentionally does "nothing", I don't think it is a problem.

PR #256, opened 2020-08-07
Extract plugin for configuration of native-platform

wolfs:


buildSrc/src/main/java/gradlebuild/JniPlugin.java
```diff
@@ -144,6 +144,7 @@ private void configureNativeJars(Project project, VariantsExtension variants, bo
         // We register the publications here, so they are available when the project is used as a composite build.
 // When we don't use the software model plugins anymore, then this can move out of the afterEvaluate block.
         project.afterEvaluate(ignored -> {
+            String artifactId = project.getTasks().named("jar", Jar.class).get().getArchiveBaseName().get();
```
lptr, line 147: Why not use the project's name?

lptr:
BTW, I think it's fine to merge this, but was wondering about the things mentioned above.

buildSrc/src/main/java/gradlebuild/JniPlugin.java
wolfs: Because we are not using the project name in `testApp`.

wolfs:
> We ended up with a lot more verbose code than we started with. I wonder if writing the plugins in Kotlin (or Groovy) would have helped with that?

I think most of the verbosity comes from using Java and not Groovy/Kotlin and/or precompiled script plugins.

>  I also don't like the fact that many of the comments and the project property to test with repo artifacts is much more hidden now, though I guess that should be handled in another way.

Yeah, the flags should probably go to one place in the plugins, where they are easily discoverable. I wonder how this can be made discoverable from the command line.

> I wonder if we should try some existing JNI plugins, like the one created by Thad's team: https://github.com/wpilibsuite/gradle-jni

We could, though I don't know what they already provide and what we would need to add.

PR #257, opened 2020-08-10
Move file events to its own subproject

wolfs:


wolfs:
> LGTM. Shall we keep calling things `file-events` or do we want to later rename it to something like `file-system-watching` I wonder...

I think the package structure in the `file-events` subproject needs to be changes as well as the artifact id. I thought the artifact id of the new library could be something like `native-file-system-watching`?

I wonder if we should also try to publish the library under the `org.gradle` groupId. That would involve some extra work currently, since I don't now how to publish to jCenter or mavenCentral under the `org.gradle` organization.

lptr:
Do you want to do these things in this PR? I think it'd be best to separate things first, and then we can do more renames.

I don't like the idea of publishing artifacts with different group IDs from the same repo.

wolfs:
> I don't like the idea of publishing artifacts with different group IDs from the same repo.

I agree that we should use the same `groupId` when we publish from this repository.

> Do you want to do these things in this PR? I think it'd be best to separate things first, and then we can do more renames.

It would be good if we could come up with a good `artifactId` and project name for the new subproject before merging this PR. I agree that we can do this in later PRs as well.

gradle/verification-metadata.xml
```diff
@@ -8,6 +8,7 @@
          <trust file=".*-sources[.]jar" regex="true"/>
          <trust file=".*-src[.]zip" regex="true"/>
          <trust group="net[.]rubygrapefruit" name="native-platform.*" regex="true"/>
+         <trust group="net[.]rubygrapefruit" name="file-events.*" regex="true"/>
```
lptr, line 11: Why don't we trust all entries in `net.rubygrapefruit`?

wolfs:
Before we merge this we should make sure we can use the new library in the Gradle build: gradle/gradle#14095.

gradle/verification-metadata.xml
```diff
@@ -7,8 +7,7 @@
          <trust file=".*-javadoc[.]jar" regex="true"/>
          <trust file=".*-sources[.]jar" regex="true"/>
          <trust file=".*-src[.]zip" regex="true"/>
-         <trust group="net[.]rubygrapefruit" name="native-platform.*" regex="true"/>
-         <trust group="net[.]rubygrapefruit" name="file-events.*" regex="true"/>
+         <trust group="net.rubygrapefruit"/>
```
lptr, line 10: Why do we not need the `[]` around the dot anymore?

wolfs: Because the `group` name isn't treated as a regex any more, since the `regex="true"` tag is gone.

PR #261, opened 2020-11-13
Report failure when watcher thread exits without being shutdown

### LINKED ISSUE #260: Report an error when the file watching thread stops
When `executeRunLoop0` throws an error in the file watcher thread, this goes unnoticed and the watching thread stops. That means that the watcher does not look for changes any more, even though it did not notify the client about this. This causes changes to go undetected, as for example shown in this Gradle issue: https://github.com/gradle/gradle/issues/15115.

wolfs:
Fixes #260


lptr (approved):
LGTM, though I wonder if at one point in the callback we'll want to distinguish between errors that caused the thread to exit and errors that did not do that.

wolfs:
> I wonder if at one point in the callback we'll want to distinguish between errors that caused the thread to exit and errors that did not do that.

We probably should distinguish between those errors at some point. This seems more about recoverable failures and non-recoverable failures, though. When the thread stops for some reason, that is a non-recoverable failure. I don't know if we have some recoverable failures reported from the native side, or if all the failures on the native side are non-recoverable as well.

Issue #262, opened 2020-12-16
MemoryFunctions.getMemoryInfo() returns negative available memory on macOS Big Sur


lptr:
With this code:

https://github.com/gradle/native-platform/blob/1bc4713133cea2ac9d6bc838900d77c6a6a28a20/native-platform%2Fsrc%2Fmain%2Fcpp%2Fapple.cpp#L139-L143

the `MemoryTest` is failing:

```text
net.rubygrapefruit.platform.memory.MemoryTest > can query system memory FAILED
    Condition not satisfied:

    memoryInfo.availablePhysicalMemory > 0
    |          |                       |
    |          -4287809867776          false
    <net.rubygrapefruit.platform.internal.DefaultMemoryInfo@1f50969f totalMem=137438953472 availableMem=-4287809867776>
        at net.rubygrapefruit.platform.memory.MemoryTest.can query system memory(MemoryTest.groovy:41)
```

I've printed the actual numbers being used in the calculation:

```text
free_count:           20605006
inactive_count:        5259661
speculative_count:  1072693248
```

Build scan: https://ge.gradle.org/s/23pj72qraz7z6/console-log?anchor=72

lptr:
```console
$ vm_stat
Mach Virtual Memory Statistics: (page size of 4096 bytes)
Pages free:                               74120.
Pages active:                           2885788.
Pages inactive:                         2876023.
Pages speculative:                        10223.
Pages throttled:                              0.
Pages wired down:                       1064847.
Pages purgeable:                         130731.
"Translation faults":                 719998571.
Pages copy-on-write:                   13875681.
Pages zero filled:                    225231206.
Pages reactivated:                     77295470.
Pages purged:                           2265126.
File-backed pages:                      1510903.
Anonymous pages:                        4261131.
Pages stored in compressor:             4394742.
Pages occupied by compressor:           1476978.
Decompressions:                        53721785.
Compressions:                          69558456.
Pageins:                               22989266.
Pageouts:                                567396.
Swapins:                                7980369.
Swapouts:                               9106971.
```

```console
$ vm_stat
Mach Virtual Memory Statistics: (page size of 4096 bytes)
Pages free:                             8028396.
Pages active:                           3679904.
Pages inactive:                         3537410.
Pages speculative:                       187251.
Pages throttled:                              0.
Pages wired down:                       1189073.
Pages purgeable:                          96555.
"Translation faults":                3801806791.
Pages copy-on-write:                  139437721.
[...]

PR #263, opened 2020-12-16
Temporarily disable failing test

lptr:
This is to work around https://github.com/gradle/native-platform/issues/262 for now.


wolfs (commented):
Looks good. Could we also disable the test on M1 only?

lptr:
Not sure how to detect M1. Any ideas?

wolfs:
Why are we ignoring the test now for all operating systems and not only for macOS?

lptr:
The test was only enabled for macOS. Now it's not enabled at all.

PR #264, opened 2020-12-16
Upgradle to 6.7.1

lptr:


gradle/verification-metadata.xml
```diff
@@ -11,17 +11,17 @@
       </trusted-artifacts>
       <trusted-keys>
          <trusted-key id="0785b3eff60b1b1bea94e0bb7c25280eae63ebe5" group="^org[.]apache[.]httpcomponents($|([.].*))" regex="true"/>
+         <trusted-key id="2db4f1ef0fa761ecc4ea935c86fdc7e2a11262cb" group="commons-codec"/>
```
wolfs, line 14: What is the reason for moving those entries around?

gradle/verification-metadata.xml
```diff
@@ -35,6 +35,31 @@
             <pgp value="7615ad56144df2376f49d98b1669c4bb543e0445"/>
          </artifact>
       </component>
+      <component group="com.google.guava" name="failureaccess" version="1.0.1">
```
wolfs, line 38: Those entries shouldn't be necessary and are covered by the accepted keys above. Is that not the case? I think adding only the conventions plugin should be enough.

wolfs (commented):
The upgrade looks good. I think we need much fewer changes in the verification metadata file.

gradle/verification-metadata.xml
lptr: I'm assuming that they are now ordered in some way.

lptr: Yeah, it looks like they are now ordered alphabetically by `id`.

gradle/verification-metadata.xml
lptr: OK. I wonder why the automated task doesn't just skip them then...

gradle/verification-metadata.xml
lptr: TESTING

gradle/verification-metadata.xml
```diff
@@ -50,9 +55,9 @@
             <pgp value="19beab2d799c020f17c69126b16698a4adf4d638"/>
          </artifact>
       </component>
-      <component group="org.gradle" name="test-retry-gradle-plugin" version="1.1.1">
-         <artifact name="test-retry-gradle-plugin-1.1.1.jar">
-            <sha256 value="3d5fc32376b92bd87f405ca649b8bab4333afc1e03954945f26ab657e0c1fedf" origin="Generated by Gradle because artifact wasn't signed"/>
+      <component group="org.gradle" name="test-retry-gradle-plugin" version="1.1.8">
```
wolfs, line 58: The test retry plugin entry isn't necessary any more, since it is now signed. Please remove it.

lptr: Removed.

wolfs (approved):
LGTM!

PR #265, opened 2020-12-16
Add support for Apple silicon

lptr:


lptr:
This now compiles locally on the M1 Mac mini, but I can't test the CI build without merging it to `master`. So I'm going to do that, we can do a review post-merge. /cc @wolfs

buildSrc/src/main/java/gradlebuild/JniPlugin.java
```diff
@@ -263,7 +264,13 @@ void createPlatforms(PlatformContainer platformContainer) {
                 // https://github.com/gradle/gradle/blob/36614ee523e5906ddfa1fed9a5dc00a5addac1b0/subprojects/platform-native/src/main/java/org/gradle/nativeplatform/toolchain/internal/gcc/AbstractGccCompatibleToolChain.java
                 toolChain.target("linux_aarch64");
             });
-            toolChainRegistry.create("clang", Clang.class);
+            toolChainRegistry.create("clang", Clang.class, toolChain -> {
+                // The core Gradle toolchain for Clang only targets x86 and x86_64 out of the box.
+                OperatingSystem os = new DefaultNativePlatform("current").getOperatingSystem();
+                if (os.isMacOsX()) {
+                    toolChain.target("osx_aarch64");
```
wolfs, line 271: Does it target `osx_arm64`? I wonder how we determine the default Clang targets. Is that some code in Gradle?

wolfs (commented):
LGTM! I left one comment.

buildSrc/src/main/java/gradlebuild/JniPlugin.java
lptr: No idea, but without this it doesn't work, and with it it does. I guess we can remove the `if` here after https://github.com/gradle/native-platform/pull/266.

wolfs: I think we can remove the `if` when we use a `7.0` Gradle milestone for building native platform itself. We should add `osx_aarch64` as target on non-macOS operating systems.

Shawyeok:
@lptr 
Nice work. BTW, Is there any plan to release 0.22? looks like 0.21 is released in Dec 2019, which is four years ago.

lptr:
Yeah, that's something we should have sorted out a long time ago. There isn't much intention in supporting the new features we've added since 0.21 (mostly about file-system watching) outside of Gradle. We should probably move those out to some other place, and release 0.22 with the few changes that have happened in the already existing features.

PR #275, opened 2021-04-13
Support building multiple ncurses variants

wolfs:
to support Ubuntu 20.

### DIFF
--- a/buildSrc/src/main/java/gradlebuild/NcursesPlugin.java
+++ b/buildSrc/src/main/java/gradlebuild/NcursesPlugin.java
@@ -71,9 +71,9 @@ public class NcursesPlugin extends RuleSource {
             return ImmutableSortedSet.of("5");
         }
         ImmutableSortedSet.Builder<String> builder = ImmutableSortedSet.naturalOrder();
-        for (String d : ImmutableList.of("/lib", "/lib64", "/lib/x86_64-linux-gnu", "/lib/aarch64-linux-gnu")) {
+        for (String d : ImmutableList.of("/lib", "/lib64", "/lib/x86_64-linux-gnu", "/lib/aarch64-linux-gnu", "/usr/lib")) {
             File libDir = new File(d);
-            if (new File(libDir, "libncurses.so.6").isFile()) {
+            if (new File(libDir, "libncurses.so.6").isFile() || new File(libDir, "libncursesw.so.6").isFile()) {
                 builder.add("6");
             }
             if (new File(libDir, "libncurses.so.5").isFile()) {
--- a/buildSrc/src/main/java/gradlebuild/NcursesPlugin.java
+++ b/buildSrc/src/main/java/gradlebuild/NcursesPlugin.java
@@ -1,7 +1,7 @@
 package gradlebuild;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSortedSet;
+import com.google.common.collect.ImmutableSet;
 import org.gradle.model.Each;
 import org.gradle.model.Model;
 import org.gradle.model.Mutate;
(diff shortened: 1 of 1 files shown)


lptr:
Please also include the changes from here: [`master...lptr:lptr/test-issue-16787`#diff-d276591c39](https://github.com/gradle/native-platform/compare/master...lptr:lptr/test-issue-16787#diff-d276591c392b8c1be3d53249633fa8ccdb74ec05abfaee181c659e5de824910c)

This allows detecting ncurses versions on Alpine Linux.

buildSrc/src/main/java/gradlebuild/NcursesPlugin.java
```diff
@@ -54,21 +65,26 @@
...
-        for (String d : ImmutableList.of("/lib", "/lib64", "/lib/x86_64-linux-gnu", "/lib/aarch64-linux-gnu")) {
+        ImmutableSortedSet.Builder<String> builder = ImmutableSortedSet.naturalOrder();
+        for (String d : ImmutableList.of("/lib", "/lib64", "/lib/x86_64-linux-gnu", "/lib/aarch64-linux-gnu", "/usr/lib")) {
             File libDir = new File(d);
-            if (new File(libDir, "libncurses.so.6").isFile()) {
-                return "6";
+            if (new File(libDir, "libncurses.so.6").isFile() || new File(libDir, "libncursesw.so.6").isFile()) {
+                builder.add("6");
             }
             if (new File(libDir, "libncurses.so.5").isFile()) {
```
lptr, line 79: Why not detect `libncursesw.so.5` as well?

buildSrc/src/main/java/gradlebuild/NcursesPlugin.java
```diff
@@ -30,21 +36,26 @@
...
-            if (targetPlatform.getOperatingSystem().isLinux() && !ncursesVersion.getNcursesVersion().equals("5")) {
+            if (targetPlatform.getOperatingSystem().isLinux() && !isNcursesVersion(targetPlatform, "5")) {
                 binarySpec.getLinker().args("-lncursesw");
             } else {
                 binarySpec.getLinker().args("-lcurses");
             }
         }
     }
+
+    private boolean isNcursesVersion(NativePlatform targetPlatform, String ncursesVersion) {
```
lptr, line 55: Nit: let's pass the `NcursesVersion` object here. It would also be nicer to rename `NcursesVersion.getNcursesVersion()` to something like `getVersionNumber()` or something less redundant.

buildSrc/src/main/java/gradlebuild/NcursesPlugin.java
```diff
@@ -14,13 +15,18 @@
...
 import static gradlebuild.NativeRulesUtils.addPlatform;
 
 public class NcursesPlugin extends RuleSource {
 
-    @Model NcursesVersion ncursesVersion() {
-        return new NcursesVersion(inferNCursesVersion());
+    @Model List<NcursesVersion> ncursesVersions() {
+        return inferNCursesVersions().stream()
+            .map(NcursesVersion::new)
+            .collect(Collectors.toList());
```
lptr, line 29: Nit: would be nicer if `inferNCursesVersions()` would already return `NcursesVersion` objects.

Converting between a `List` and a `SortedSet` also looks unnecessary. Can we just merge these?

lptr (commented):
I left one question and a few nits.

buildSrc/src/main/java/gradlebuild/NcursesPlugin.java
wolfs: Because there doesn't seem to be a `w` variant for ncurses5: If you check above we link against `ncursesw` for ncurses 6 on Linux and against `ncurses` for everything else.

buildSrc/src/main/java/gradlebuild/NcursesPlugin.java
wolfs: I changed the usage of `NcursesVersion`. PTAL!

lptr (approved):
LGTM.

PR #277, opened 2021-04-14
Handle gracefully if fs details can't be read

wolfs:
This seems to happen on Windows for network file systems.


wolfs:
I tried out if the calls to `addUnknown` actually succeed by calling this method from native calls: https://builds.gradle.org/buildConfiguration/GradleNative_NativePlatform_BuildTrigger/42592390?buildTab=dependencies&mode=timeline&range=1618410033000-1618413963000

Seems to work.

native-platform/src/main/cpp/apple.cpp
```diff
@@ -71,27 +72,27 @@ Java_net_rubygrapefruit_platform_internal_jni_PosixFileSystemFunctions_listFileS
...
-        }
-
-        if (alist.volattr & ATTR_VOL_CAPABILITIES) {
-            if ((buffer.caps.valid[VOL_CAPABILITIES_FORMAT] & VOL_CAP_FMT_CASE_SENSITIVE)) {
-                caseSensitive = (buffer.caps.capabilities[VOL_CAPABILITIES_FORMAT] & VOL_CAP_FMT_CASE_SENSITIVE) != 0;
+            env->CallVoidMethod(info, unknownFsMethod, mount_point, file_system_type, device_name, remote);
+        } else {
+            if (alist.volattr & ATTR_VOL_CAPABILITIES) {
+                if ((buffer.caps.valid[VOL_CAPABILITIES_FORMAT] & VOL_CAP_FMT_CASE_SENSITIVE)) {
+                    caseSensitive = (buffer.caps.capabilities[VOL_CAPABILITIES_FORMAT] & VOL_CAP_FMT_CASE_SENSITIVE) != 0;
```
lptr, line 87: Nt: let's move the definition of `caseSensitive` and `casePreserving` here.

native-platform/src/main/java/net/rubygrapefruit/platform/file/FileSystemInfo.java
```diff
@@ -37,6 +37,20 @@
...
+     * You must not query additional details if they are unknown.
+     * <p>
+     * Additional details are:
+     * <ul>
+     *     <li>{@link #isCasePreserving()}</li>
+     *     <li>{@link #isCaseSensitive()}</li>
+     * </ul>
+     */
+    @ThreadSafe
+    boolean isDetailsKnown();
```
lptr, line 52: As this is part of the public API, I think it's important to have nice names, and `isDetailsKnown()` has a singular/plural collision. Maybe `isCaseSensitivityKnown()` would be a way to work around it?

lptr (commented):
I'm a bit uneasy about the resulting API throwing exceptions when you access the `FileSystemInfo` instance.

I'd rather keep the current API fail, and introduce a new API to query file systems in a way that details are returned as `Optional` or something. WDYT?

native-platform/src/main/cpp/apple.cpp
wolfs: Done.

wolfs:
@lptr I changed how we handle the case when we can't determine the case sensitivity. The API is now a little bit different, though I think it is better that way. We can't use `Optional`, since we are still on Java 6, so I used a nullable method instead.

lptr (approved):
I like this solution, LGTM!

PR #278, opened 2021-04-20
Fix bug for resolving file system type on Windows

wolfs:
We converted the C-String to a Java String before `GetVolumeInformationW` wrote it.

### DIFF
--- a/native-platform/src/main/cpp/win.cpp
+++ b/native-platform/src/main/cpp/win.cpp
@@ -355,7 +355,6 @@ Java_net_rubygrapefruit_platform_internal_jni_PosixFileSystemFunctions_listFileS
             }
 
             jstring mount_point = wchar_to_java(env, cur, wcslen(cur), result);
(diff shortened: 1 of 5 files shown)


native-platform/src/main/cpp/win.cpp
```diff
@@ -364,7 +363,7 @@ Java_net_rubygrapefruit_platform_internal_jni_PosixFileSystemFunctions_listFileS
                 if (GetVolumeInformationW(cur, NULL, 0, NULL, NULL, &flags, fileSystemName, MAX_PATH + 1) == 0) {
                     env->CallVoidMethod(info, unknownFsMethod,
                         mount_point,
-                        file_system_type,
+                        NULL,
```
lptr, line 366: I think I'd use something like `unknown` here instead to avoid having to deal with `null`. WDYT?

wolfs: I am using `unknown` now, though I deal with the problem on the Java side and continue to return `null` here, so I don't have to deal with strings in C.

PR #280, opened 2021-04-30
Ignore already removed watch points on Linux

lptr:
On Alpine Linux in a Docker container we've seen that it is possible that calling `inotify_rm_watch()` immediately produces the `IN_IGNORED` event, which in turn can result in the watch point to be deleted before returning from the `cancel()` method, resulting in a crash.

This change ensures that we ignore already removed watch points instead.

### DIFF
--- a/file-events/src/file-events/cpp/linux_fsnotifier.cpp
+++ b/file-events/src/file-events/cpp/linux_fsnotifier.cpp
@@ -298,11 +298,10 @@ void Server::registerPath(const u16string& path) {
     if (watchRoots.find(watchDescriptor) != watchRoots.end()) {
         throw FileWatcherException("Already watching path", path);
     }
-    auto result = watchPoints.emplace(piecewise_construct,
+    watchPoints.emplace(piecewise_construct,
         forward_as_tuple(path),
         forward_as_tuple(path, inotify, watchDescriptor));
-    auto& watchPoint = result.first->second;
-    watchRoots[watchPoint.watchDescriptor] = path;
+    watchRoots[watchDescriptor] = path;
 }
 
 bool Server::unregisterPath(const u16string& path) {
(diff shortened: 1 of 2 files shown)


wolfs (approved):
LGTM!

file-events/src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -298,11 +298,10 @@ void Server::registerPath(const u16string& path) {
     if (watchRoots.find(watchDescriptor) != watchRoots.end()) {
         throw FileWatcherException("Already watching path", path);
     }
-    auto result = watchPoints.emplace(piecewise_construct,
+    watchPoints.emplace(piecewise_construct,
```
wolfs, line 301: I think we should add a comment here, why we don't re-use the return value of emplace.

file-events/src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -312,13 +311,14 @@ bool Server::unregisterPath(const u16string& path) {
...
     auto& watchPoint = it->second;
+    int wd = watchPoint.watchDescriptor;
     CancelResult ret = watchPoint.cancel();
     if (ret == CancelResult::ALREADY_CANCELLED) {
         return false;
     }
-    recentlyUnregisteredWatchRoots.emplace(watchPoint.watchDescriptor, path);
-    watchRoots.erase(watchPoint.watchDescriptor);
-    watchPoints.erase(it);
+    recentlyUnregisteredWatchRoots.emplace(wd, path);
```
wolfs, line 319: I think we should add a comment here why we don't use the value of `watchPoint.watchDescriptor` any more.

wolfs (commented):
I think we should add some comments so we don't accidentally introduce the problem again.

file-events/src/file-events/cpp/linux_fsnotifier.cpp
lptr: I don't think that's necessary. IIUC, it's quite non-typical to do so actually.

file-events/src/file-events/cpp/linux_fsnotifier.cpp
lptr: I wanted to add that comment about how `cancel()` might already trigger the event, but it doesn't check out. This is basically defensive coding at this point, but I don't have a good rationalization why it's needed.

wolfs: Yeah, adding a comment about being defensive here because if we aren't we are going to get a crash on Alpine would be good enough, even if we don't understand the reason. To make sure we don't accidentally re-introduce the problem when we refactor the code.

lptr: This is not the cause of the crash; it was caused by the iterator going missing. I added a comment here: https://github.com/gradle/native-platform/pull/282

PR #283, opened 2021-05-07
Disable file events on non-glibc implementations on Linux

lptr:
This is primarily targeted at musl libc that seems to have problems with watching files on Linux, causing crashes in Gradle. See https://github.com/gradle/gradle/issues/17099.

### DIFF
--- a/file-events/src/file-events/cpp/linux_fsnotifier.cpp
+++ b/file-events/src/file-events/cpp/linux_fsnotifier.cpp
@@ -1,6 +1,7 @@
 #ifdef __linux__
 
 #include <codecvt>
+#include <dlfcn.h>
 #include <locale>
 #include <string>
 #include <sys/ioctl.h>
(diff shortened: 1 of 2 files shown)


file-events/src/main/java/net/rubygrapefruit/platform/internal/jni/AbstractFileEventFunctions.java
```diff
@@ -22,11 +22,28 @@ public static String getVersion() {
...
+    }
+
     /**
      * Forces the native backend to drop the cached JUL log level and thus
      * re-query it the next time it tries to log something to the Java side.
+     *
+     * @throws NativeException if file events are not supported.
      */
     public void invalidateLogLevelCache() {
+        ensureSupported();
```
wolfs, line 46: I think this should be better done in `FileEvents.getEventFunctions()`, so you can't even get the functions when glibc is not supported.

lptr: Good point.

wolfs (approved):
LGTM!

PR #284, opened 2021-05-11
Throw NativeIntegrationUnavailableException when on musl libc

lptr:
We shouldn't throw `NativeException`, but `NativeIntegrationUnavailableException` when the integration is not available.


wolfs (approved):
LGTM!

PR #285, opened 2021-05-14
Sign artifacts

### LINKED ISSUE #267: Sign released artifacts
So that they can be consumed easily with dependency verification.

wolfs:
and remove publishing to bintray, since that isn't possible any more.

Fixes #267


wolfs:
Using the singed snapshot version adding `net.rubygrapefruit` to the list of trusted artifacts [here](https://github.com/gradle/gradle/blob/ab3fc9e630532d7115c63bffe58b65abfb3df9fa/gradle/verification-metadata.xml#L79-L84) seems to work:
```
         <trusted-key id="314fe82e5a4c5377bca2edec5208812e1e4a6db0">
            <trusting group="com.gradle"/>
            <trusting group="org.gradle"/>
            <trusting group="net.rubygrapefruit"/>
            <trusting group="^com[.]gradle($|([.].*))" regex="true"/>
            <trusting group="com.gradle" name="gradle-enterprise-gradle-plugin"/>
         </trusted-key>
```

Note that I removed all the `component` entries for native platform and the `<trust group="net[.]rubygrapefruit" name="(native-platform|file-events).*" version=".*snapshot.*" regex="true"/>` entry to test this.

buildSrc/src/main/java/gradlebuild/PublishPlugin.java
```diff
@@ -3,31 +3,122 @@
...
+        TaskContainer tasks = project.getTasks();
+        TaskProvider<Jar> jarTask = project.getTasks().named("jar", Jar.class);
+        TaskProvider<Jar> sourceZipTask = tasks.register("sourceZip", Jar.class, jar -> jar.getArchiveClassifier().set("sources"));
+        TaskProvider<Jar> javadocZipTask = tasks.register("javadocZip", Jar.class, jar -> jar.getArchiveClassifier().set("javadoc"));
+        project.getExtensions().configure(
+            PublishingExtension.class,
+            extension -> extension.getPublications().create("main", MavenPublication.class, main -> {
+                main.artifact(jarTask.get());
+                main.artifact(sourceZipTask.get());
+                main.artifact(javadocZipTask.get());
```
lptr, line 49: Do you need to call `get()` here? Can't we just pass the provider?

buildSrc/src/main/java/gradlebuild/PublishPlugin.java
```diff
@@ -3,31 +3,122 @@
...
+            }));
+        project.afterEvaluate(ignored -> project.getExtensions().configure(
+            PublishingExtension.class,
+            extension -> extension.getPublications().named("main", MavenPublication.class, main -> {
+                main.setGroupId(project.getGroup().toString());
+                main.setArtifactId(jarTask.get().getArchiveBaseName().get());
+                main.setVersion(project.getVersion().toString());
```
lptr, line 56: These setters should accept `Provider<String>` and then we could avoid having to use `afterEvaluate()`, right?

buildSrc/src/main/java/gradlebuild/PublishPlugin.java
```diff
@@ -3,31 +3,122 @@
...
+            })));
```
lptr, line 57: I wonder if it would make sense to add some comments about why we need to split the two parts of configuring this extension the way it's done now.

buildSrc/src/main/java/gradlebuild/PublishPlugin.java
```diff
@@ -3,31 +3,122 @@
...
+        if (project.hasProperty(PUBLISH_API_KEY_PROPERTY)) {
+            credentials.setApiKey(project.property(PUBLISH_API_KEY_PROPERTY).toString());
+        }
+        return credentials;
+    }
+
+    public static String getArchivesBaseName(Project project) {
+        BasePluginConvention convention = project.getConvention().findPlugin(BasePluginConvention.class);
+        return convention != null ? convention.getArchivesBaseName() : project.getName();
+    }
```
lptr, line 105: When is `convention` `null` here? Will this be better with `BasePluginExtension` being introduced with Gradle 7.1?

buildSrc/src/main/java/gradlebuild/PublishPlugin.java
wolfs: I think so.

buildSrc/src/main/java/gradlebuild/PublishPlugin.java
wolfs: I think the `root` project does not apply the `BasePlugin`, that is when the convention will be null. I don't remember why I introduced it back then, I suppose it was necessary. I don't know if this is still necessary though.

buildSrc/src/main/java/gradlebuild/PublishPlugin.java
wolfs: Looks like the provider is supported starting from Gradle 6.6: https://github.com/gradle/gradle/pull/13505

Let me try to remove the `get` now.

wolfs: Removed the `get()`.

buildSrc/src/main/java/gradlebuild/PublishPlugin.java
wolfs: I added a comment.

buildSrc/src/main/java/gradlebuild/PublishPlugin.java
wolfs: I removed the `null` handling. I don't know about `BasePluginExtension` or what would be different there.

lptr (approved):
LGTM.

PR #287, opened 2021-06-08
Use new ReadDirectoryChangesExW() API

lptr:
Use the new `ReadDirectoryChangesExW()` API (instead of the old `ReadDirectoryChangesW()`) introduced in Windows 10, version 1907 in 2017.

The new API is available on all Windows 10 versions currently supported by Microsoft except for two LTSC versions released in 2015 and 2016. The latest LTSC version released in 2018 is supported.

On unsupported versions of Windows the expected behavior is that the library will fail to load with a missing symbol, and thus we'll report that file events are unavailable.

The new API does not seem to emit as many random extra events (aka "optional events"), though this might be something that was also fixed in the old API in a recent Windows update. In any case, the PR removes most (but not all) `optionalChange()` expectations.

### DIFF
--- a/file-events/src/file-events/cpp/win_fsnotifier.cpp
+++ b/file-events/src/file-events/cpp/win_fsnotifier.cpp
@@ -84,7 +84,7 @@ bool WatchPoint::isValidDirectory() {
 }
 
 ListenResult WatchPoint::listen() {
-    BOOL success = ReadDirectoryChangesW(
+    BOOL success = ReadDirectoryChangesExW(
         directoryHandle,              // handle to directory
         &buffer[0],                   // read results buffer
         (DWORD) buffer.capacity(),    // length of buffer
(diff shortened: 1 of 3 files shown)


lptr:
Related Gradle issue: https://github.com/gradle/gradle/issues/12756

lptr:
We could use an old image from here to test the unsupported behavior: https://github.com/boxcutter/windows (thanks @donat)

file-events/src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -84,15 +84,16 @@ bool WatchPoint::isValidDirectory() {
 }
 
 ListenResult WatchPoint::listen() {
-    BOOL success = ReadDirectoryChangesW(
+    BOOL success = ReadDirectoryChangesExW(
         directoryHandle,              // handle to directory
         &buffer[0],                   // read results buffer
```
wolfs, line 89: Do you think we need to increase the buffer size, now that the data structure is a little bit bigger?

wolfs (approved):
Looks good! I left one small comment. Please report the results of the manual tests on older Windows versions on the PR.

file-events/src/file-events/cpp/win_fsnotifier.cpp
lptr: It's just a default, and I have no good idea about how to size it properly, so I'd leave it at the current value. We reconfigure it to 64K on the Gradle side anyway, because that's the maximum supported on network drives.

PR #289, opened 2021-07-23
Report files changed via short paths as long paths on Windows

lptr:
When a file (e.g. `test-file.txt`) is changed via its short name (`TEST-F~1.TXT`) instead of its long name, the notification also arrives using the short name. Actually, the WINAPI docs say it's not even sure which:

> If there is both a short and long name for the file, the function will return one of these names, but it is unspecified which one.

https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-file_notify_extended_information#:~:text=character.-,filename,-A

This can be confusing for clients that keep track of the long names. This PR adds a way to enforce a consistent behavior where changes are always reported against the long path.

### DIFF
--- a/file-events/src/file-events/cpp/jni_support.cpp
+++ b/file-events/src/file-events/cpp/jni_support.cpp
@@ -110,19 +110,6 @@ void javaToUtf16StringArray(JNIEnv* env, jobjectArray javaStrings, vector<u16str
     }
 }
 
-// Utility wrapper to adapt locale-bound facets for wstring convert
-// Exposes the protected destructor as public
-// See https://en.cppreference.com/w/cpp/locale/codecvt
-template <class Facet>
-struct deletable_facet : Facet {
-    template <class... Args>
-    deletable_facet(Args&&... args)
-        : Facet(forward<Args>(args)...) {
-    }
-    ~deletable_facet() {
-    }
-};
-
 u16string utf8ToUtf16String(const char* string) {
     wstring_convert<deletable_facet<codecvt<char16_t, char, mbstate_t>>, char16_t> conv16;
     return conv16.from_bytes(string);
--- a/file-events/src/file-events/cpp/win_fsnotifier.cpp
+++ b/file-events/src/file-events/cpp/win_fsnotifier.cpp
@@ -9,7 +9,7 @@ using namespace std;
 // WatchPoint
 //
 
-WatchPoint::WatchPoint(Server* server, size_t bufferSize, const wstring& path)
+WatchPoint::WatchPoint(Server* server, size_t eventBufferSize, const wstring& path)
     : pathW(path)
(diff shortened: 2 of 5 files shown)


lptr:
The idea is to use `GetLongPathNameW()` to resolve the long path for every file change. This will make the process somewhat slower.

lptr:
Looks like `GetLongPathNameW()` works well for created and changed files, but not for deletions and removals.

```text
org.spockframework.util.InternalSpockError: Events received differ from expected:
 - UNEXPECTED FAILURE net.rubygrapefruit.platform.NativeException: Cannot convert changed path to long path, error = 2: C:\Users\vmadmin\native-platform\file-events\build\test files\junit5203341711907204467\can detect file removed\root\removed.txt
 - MISSING    MODIFIED ...\root\removed.txt
 - MISSING    REMOVED ...\root\removed.txt
```

I guess what we can do is if we encounter a `ERROR_FILE_NOT_FOUND` error, we could either ignore it, or we could try to resolve the parent directory's long path, and graft the (short) file name back to the result. In theory a directory must be emptied first before it can be removed, so we _should_ always have a chance to look up the parent dir...

Another option is that we seem to always receive a `MODIFIED` event together with the `REMOVED` event. Maybe we can look up the long path after the `MODIFIED` event, and use it to process the `REMOVED` event... That is assuming they come in that order, and that by the time we are processing the `MODIFIED` event `GetLongPathNameW()` still has a chance to access the file...

lptr:
Looks like the event order is good:

```text
net.rubygrapefruit.platform.file.BasicFileEventFunctionsTest > can detect file removed STANDARD_ERROR
    20:45:17.021  INFO    >>> Running 'can detect file removed'
    20:45:17.049  INFO    > Creating ...\root\removed.txt
    20:45:17.051  INFO    < Created ...\root\removed.txt
    20:45:17.141  INFO    > Expecting MODIFIED ...\root\removed.txt
    20:45:17.142  INFO    > Expecting REMOVED ...\root\removed.txt
    20:45:17.175  INFO    > Received MODIFIED C:\Users\vmadmin\native-platform\file-events\build\test files\junit2126691140238001409\can detect file removed\root\removed.txt
    20:45:17.195  INFO    > Received REMOVED C:\Users\vmadmin\native-platform\file-events\build\test files\junit2126691140238001409\can detect file removed\root\removed.txt
    20:45:17.225  INFO    <<< Finished 'can detect file removed'
```

lptr:
...but of course we get both events at once, when the file is already gone, so we can't call `GetLongPathNameW()` anymore. :/

So that leaves us with #1. If we can't look up the parent either, then we should be able to conclude that the parent was removed, too, and we will receive a separate notification for that, too, so we can simply ignore the current notification...

lptr:
There is no good way to handle deleted content, so let's not do this.

PR #290, opened 2021-07-29
Handle watched directory moved on Windows

lptr:
Windows handles file watching using relative paths. When the watched directory is moved, Windows sends no notification. A client keeping track of absolute paths faces two problems:

1) when changes are made to the contents of the moved directory, changes are still reported as if they happened under the original, registered root,
2) there is no way to know if the original location has completely disappeared due to a move.

(Note that if the original watched root is deleted, Windows does send a notification, so this is just about renames/moves.)

The problems are addressed in this PR as follows:

1) we are now reporting the moved locations for further changes,
2) while we still can't get notified about the original location being moved away, there is now a new API (`WindowsFileWatcher.stopWatchingMovedLocations()`) that can be used to stop watching any locations that have been moved since registration. The new API returns the _originally registered path_ of all stopped watch points.

The PR also includes the following changes:

- it exposes platform-specific `FileWatcher` implementations to allow for the Windows-specific functionality to be accessed,
- Windows paths are now handled as `wstring`s,
- we keep track of the registered watchers as a `std::list` instead of `std::unordered_map` to reduce potentially stale path references. (We still keep the originally registered path in the `WatchPoint` instance.)

### DIFF
--- a/file-events/src/file-events/cpp/win_fsnotifier.cpp
+++ b/file-events/src/file-events/cpp/win_fsnotifier.cpp
@@ -15,6 +15,67 @@ string wideToUtf8String(const wstring& string) {
 
 #define wideToUtf16String(string) (u16string((string).begin(), (string).end()))
 
+bool isAbsoluteLocalPath(const wstring& path) {
+    if (path.length() < 3) {
+        return false;
+    }
+    return ((L'a' <= path[0] && path[0] <= L'z') || (L'A' <= path[0] && path[0] <= L'Z'))
+        && path[1] == L':'
+        && path[2] == L'\\';
+}
+
+bool isAbsoluteUncPath(const wstring& path) {
+    if (path.length() < 3) {
+        return false;
+    }
+    return path[0] == L'\\' && path[1] == L'\\';
+}
+
+bool isLongPath(const wstring& path) {
+    return path.length() >= 4 && path.substr(0, 4) == L"\\\\?\\";
+}
+
+bool isUncLongPath(const wstring& path) {
+    return path.length() >= 8 && path.substr(0, 8) == L"\\\\?\\UNC\\";
+}
+
+// TODO How can this be done nicer, wihtout both unnecessary copy and in-place mutation?
+void convertToLongPathIfNeeded(wstring& path) {
+    // Technically, this should be MAX_PATH (i.e. 260), except some Win32 API related
+    // to working with directory paths are actually limited to 240. It is just
+    // safer/simpler to cover both cases in one code path.
+    if (path.length() <= 240) {
+        return;
+    }
+
+    // It is already a long path, nothing to do here
+    if (isLongPath(path)) {
+        return;
+    }
+
+    if (isAbsoluteLocalPath(path)) {
+        // Format: C:\... -> \\?\C:\...
+        path.insert(0, L"\\\\?\\");
+    } else if (isAbsoluteUncPath(path)) {
+        // In this case, we need to skip the first 2 characters:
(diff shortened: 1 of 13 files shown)


file-events/src/file-events/headers/win_fsnotifier.h
```diff
@@ -59,22 +60,27 @@ enum class WatchPointStatus {
 
 class WatchPoint {
 public:
-    WatchPoint(Server* server, size_t bufferSize, const u16string& path);
+    WatchPoint(Server* server, size_t eventBufferSize, const wstring& path);
```
wolfs, line 63: I saw that you converted the string we hold to `wstring`s. What is the advantage of this?

file-events/src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -66,8 +135,25 @@ WatchPoint::~WatchPoint() {
...
+    }
+}
+
+// Allocate maximum path length
+#define PATH_BUFFER_SIZE 32768
+
+wstring WatchPoint::getPath() {
+    vector<wchar_t> buffer;
+    buffer.reserve(PATH_BUFFER_SIZE);
+    DWORD pathLength = GetFinalPathNameByHandleW(
```
wolfs, line 148: I suppose there isn't a call which doesn't follow symlinks, since the file handle already points to the target of the symlink, right?

file-events/src/file-events/cpp/win_fsnotifier.cpp
```diff
@@ -66,8 +135,25 @@ WatchPoint::~WatchPoint() {
...
+        logToJava(LogLevel::WARNING, "Couldn't cancel watch point %s: %s", wideToUtf8String(registeredPath).c_str(), ex.what());
+    }
+}
+
+// Allocate maximum path length
+#define PATH_BUFFER_SIZE 32768
+
+wstring WatchPoint::getPath() {
+    vector<wchar_t> buffer;
+    buffer.reserve(PATH_BUFFER_SIZE);
```
wolfs, line 147: Does it matter that we reserve a new buffer each time we call this method or would it be better to reserve the buffer once for the server object and re-use it?

wolfs (approved):
LGTM!

file-events/src/file-events/cpp/win_fsnotifier.cpp
lptr: I'm not sure about the reason, but yours sounds good. And indeed, I haven't found such a call. But as you can see from the changes to the tests, the behavior is now more like macOS/Linux, so I think it's fine. It doesn't matter for Gradle in any case, as we resolve the canonical file on the caller side when we register watchers.

file-events/src/file-events/cpp/win_fsnotifier.cpp
lptr: I first created the buffer for the server object, but them moved the method to the `WatchPoint` object, and I think this is just fine. It does not have to be super-efficient, and it's not like we are increasing GC time with unnecessary allocations. It's also worth keeping in mind that these operations are a result of (slow) file operations, so it's okay not to process them with the highest possible speed.

file-events/src/file-events/headers/win_fsnotifier.h
lptr: With the new changes we are converting back and forth between long/short paths a bunch more, and doing all that arithmetic in `wstring` land means we don't have to also convert back-and-forth between `u16string` and `wstring` so much.

PR #292, opened 2021-08-04
Handle moved watched handles on Windows more resiliently

lptr:
There is a bug where we re-check already `FINISHED` watch points in `stopWatchingMovedPaths()`. We should also handle the case when `getPath()` fails for whatever reason.

### DIFF
--- a/file-events/src/file-events/headers/win_fsnotifier.h
+++ b/file-events/src/file-events/headers/win_fsnotifier.h
@@ -66,10 +66,10 @@ public:
     ListenResult listen();
     bool cancel();
     /**
-     * Returns the path that is being watched. If the watched handle has been moved,
-     * this returns the new path.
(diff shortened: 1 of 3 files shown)


file-events/src/file-events/headers/win_fsnotifier.h
```diff
@@ -66,10 +66,10 @@ class WatchPoint {
     ListenResult listen();
     bool cancel();
     /**
-     * Returns the path that is being watched. If the watched handle has been moved,
-     * this returns the new path.
+     * Gets the path currently associated with the watched handle. Returns false
+     * if it failed to set the path becuase the handle was somehow inaccessible.
```
wolfs, line 70:
```suggestion
     * if it failed to set the path because the handle was somehow inaccessible.
```

Nit

wolfs (approved):
LGTM!

PR #293, opened 2021-08-04
Stop delivering events when watched root is moved on Windows

lptr:
OK, so we need to handle `SUBST` drives properly, too... New plan. The watched root can be referred to via a bunch of different locations:

- the **registered path** is a string we get from the Java side that might point to a symlink, use a different casing than the actual target file, or point to a location on a SUBST drive. We use this path to identify registered watch points to the Java side. (Note that if you register a watched by passing it a `File.getCanonicalFile()`, it _will_ resolve symlinks and file casing, but it will _not_ resolve `SUBST` drives).
- the **registered final path** is the canonical (=final) path according to [GetFinalPathNameByHandleW](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfinalpathnamebyhandlew) present at registration, i.e. this is where the registered watch point points to on the file system. This resolves the same things as Java's `File.getCanonicalFile()`, plus it also resolves `SUBST` drives.
- the **actual final path** is the final path at the moment of either handling an event, or when checking to `stopWatchingMovedPaths()`.

The rules are the following:

- Changes are reported against the **registered path**, this means that
  - if you are watching `G:\watched`, where `G:` is a `SUBST` drive pointing to `C:\subst-target`, changes will be reported against `G:\watched and not `C:\subst-target`,
  - if you are watching `C:\LOWERCASE` even though the canonical path is `C:\lowercase`, you will get notifications against `C:\LOWERCASE`.
- If your watched directory is moved, nothing will happen until either of the following also happens:
  - If you call `WindowsFileWatcher.stopWatchingMovePaths()` first, then your moved location will be stopped.
  - If an event happens inside the moved directory first, then we'll send a `REMOVED` event against the **registered path**, _and_ stop watching. (We do this by checking the **current final path** against the **registered final path** of the watched before reporting events, and if we see it moved, we conclude that it has been deleted/moved.)
  - If the directory is moved back to its original location before an event is received of `stopWatchingMovePaths()` is called, event delivery will continue undisturbed.

### DIFF
--- a/file-events/src/test/groovy/net/rubygrapefruit/platform/file/WindowsFileEventFunctionsTest.groovy
+++ b/file-events/src/test/groovy/net/rubygrapefruit/platform/file/WindowsFileEventFunctionsTest.groovy
@@ -93,7 +93,7 @@ class WindowsFileEventFunctionsTest extends AbstractFileEventFunctionsTest {
         droppedPaths == []
     }
 
-    def "does not drop subst drive as"() {
+    def "does not drop subst drive as moved"() {
         given:
         subst("G:", rootDir)
         def watchedDir = new File("G:\\watched")


file-events/src/test/groovy/net/rubygrapefruit/platform/file/WindowsFileEventFunctionsTest.groovy
```diff
@@ -0,0 +1,121 @@
...
+        then:
+        expectNoEvents()
+
+        when:
+        droppedPaths = watcher.stopWatchingMovedPaths()
+        then:
+        droppedPaths == []
+    }
+
+    def "does not drop subst drive as"() {
```
wolfs, line 96: Should there be something after the `as`?

wolfs (commented):
I think this looks good! One small comment.

PR #296, opened 2021-09-28
Add tests for hard links

lptr:


file-events/src/test/groovy/net/rubygrapefruit/platform/file/HardLinkFileEventFunctionsTest.groovy
```diff
@@ -0,0 +1,99 @@
...
+
+@Requires({ Platform.current().macOs || Platform.current().linux || Platform.current().windows })
+class HardLinkFileEventFunctionsTest extends AbstractFileEventFunctionsTest {
+
+    def "can detect hard link created"() {
+        given:
+        def target = new File(rootDir, "target")
+        target.createNewFile()
+        def link = new File(rootDir, "link")
+        startWatcher(rootDir)
```
wolfs, line 39: Should we also have tests where the target of the hard link is outside of the watched directory?

file-events/src/test/groovy/net/rubygrapefruit/platform/file/HardLinkFileEventFunctionsTest.groovy
```diff
@@ -0,0 +1,99 @@
...
+
+        when:
+        link.text = "modified"
+
+        then:
+        // On Windows we seem to be getting multiple MODIFIED events
+        // On Linux we _sometimes_ seem to be getting multiple events
+        expectEvents byPlatform(
+            (MAC_OS): [change(MODIFIED, link)],
+            (WINDOWS): [change(MODIFIED, link)] * 2,
```
wolfs, line 70: I suppose that is one modified event for each of the watched files: one for link and one for target.

file-events/src/test/groovy/net/rubygrapefruit/platform/file/HardLinkFileEventFunctionsTest.groovy
```diff
@@ -0,0 +1,99 @@
...
+        startWatcher(rootDir)
+
+        when:
+        link.text = "modified"
+
+        then:
+        // On Windows we seem to be getting multiple MODIFIED events
+        // On Linux we _sometimes_ seem to be getting multiple events
+        expectEvents byPlatform(
+            (MAC_OS): [change(MODIFIED, link)],
```
wolfs, line 69: This and the linux behaviour is bad: If you modify the target outside of the watched directory, then we are not getting an event, right?

file-events/src/test/groovy/net/rubygrapefruit/platform/file/HardLinkFileEventFunctionsTest.groovy
lptr: Just added, it's not nice. :/

file-events/src/test/groovy/net/rubygrapefruit/platform/file/HardLinkFileEventFunctionsTest.groovy
lptr: Perhaps yes. We could try with more than two links, though whatever we'd learn doesn't sound useful.

file-events/src/test/groovy/net/rubygrapefruit/platform/file/HardLinkFileEventFunctionsTest.groovy
lptr: Actually, even on Windows you get no notification. :/

file-events/src/test/groovy/net/rubygrapefruit/platform/file/HardLinkFileEventFunctionsTest.groovy
```diff
@@ -0,0 +1,188 @@
...
+
+        then:
+        // Windows sometimes reports a modification before the removal
+        expectEvents byPlatform(
+            (WINDOWS): [optionalChange(MODIFIED, link), change(REMOVED, link)],
+            (OTHERWISE): [change(REMOVED, link)]
+        )
+    }
+
+    def "does not detect hard link removed outside watched hierarchy"() {
```
wolfs, line 166: Same here: If I remove a link pointing to a file in the watched directory, why would that even matter? Is this just for coverage so we don't see any strange events?

wolfs (approved):
This looks "good", in the sense that it tests what is currently happening. At least on Linux, I would have expected to obtain the modification event, though seems like no dice...

file-events/src/test/groovy/net/rubygrapefruit/platform/file/HardLinkFileEventFunctionsTest.groovy
lptr: Exactly, just as above.

Issue #297, opened 2021-11-27
Migrate from core native support to Nokee plugins


lacasseio:
This issue will track any work required to migrate from core Software Model plugins to the Nokee plugins for the migration pull request.

lacasseio:
Here are some of the work left to do on the PR:
- [ ] Declare Windows distribution axis.
- [ ] Declare Ncurses version axis.
- [ ] Wire Ncurse sources to the variants in `native-platform`.
- [ ] Convert FreeBSD plugin to Nokee.
- [ ] Convert Ncurses plugin to Nokee.
- [ ] Migrate JNI unit test to Nokee.

ov7a:
Given 
* https://github.com/gradle/native-platform/pull/304
* https://github.com/gradle/native-platform/pull/308
* https://github.com/gradle/native-platform/pull/316

and the fact that all of these PRs lived in a [separate branch](https://github.com/gradle/native-platform/tree/nokee-migration) for 3 years without any updates

@big-guy do we still want to proceed with that migration?

ov7a:
This effort is no longer planned, closing the issue as obsolete

PR #298, opened 2021-11-27
Migrate from core native to Nokee plugins

lacasseio:
Fixes https://github.com/gradle/native-platform/issues/297.

See the attached issues for general work items left to do. Feel free to provide any kind of feedback on this PR either by starting a new conversation or contributing to a current conversation.

### DIFF
--- a/buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
+++ b/buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
@@ -19,6 +19,7 @@ import org.gradle.model.Mutate;
 import org.gradle.model.RuleSource;
 import org.gradle.nativeplatform.platform.OperatingSystem;
 import org.gradle.nativeplatform.platform.internal.DefaultNativePlatform;
+import org.gradle.nativeplatform.tasks.LinkSharedLibrary;
 import org.gradle.nativeplatform.toolchain.Clang;
 import org.gradle.nativeplatform.toolchain.Gcc;
 import org.gradle.nativeplatform.toolchain.NativeToolChainRegistry;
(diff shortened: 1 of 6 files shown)


lacasseio:
@big-guy At the moment, the branch is using a local version so the CI will fail. I will cut a release for you once our CI stabilizes.

buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
```diff
@@ -0,0 +1,204 @@
...
+import org.gradle.nativeplatform.toolchain.Clang;
+import org.gradle.nativeplatform.toolchain.Gcc;
+import org.gradle.nativeplatform.toolchain.NativeToolChainRegistry;
+import org.gradle.nativeplatform.toolchain.VisualCpp;
+
+import java.io.File;
+import java.util.Set;
+
+@SuppressWarnings("UnstableApiUsage")
+public abstract class JniNokeePlugin implements Plugin<Project> {
```
lacasseio, line 34: ❓ Is there a reason the JNI tests are run as part of another `Test` task? Nokee wires everything for the main `Test` task. There wasn't any consideration to "remove" the wiring of the JNI into the main test task. We can create a new test task but all the wiring will have to be done manually. Nokee could provide a `jni-unit-test` plugin that allows users to declare any number of JNI test tasks but that is a bit outside of the current scope. The project could remove the extra test task in favour of the main test task however there may be a reason for the task to existing in the first place.

lacasseio:
@big-guy The dependency verification is quite unhelpful. Locally, it fails with an `Invalid UTF-8` error which is not clear as to what's the problem. I removed the verification files locally until we can work around this issue.

buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
```diff
@@ -0,0 +1,213 @@
...
+                project.getGroup().toString().replace('.', '/'),
+                "platform",
+                toVariantName(variant.getTargetMachine())));
+        });
+    }
+
+    private void registerWindowsDistributionDimension(JavaNativeInterfaceLibrary library) {
+        SetProperty<WindowsDistribution> newDimension = library.getDimensions().newAxis(WindowsDistribution.class, builder -> builder.onlyOn(library.getMachines().getWindows().getOperatingSystemFamily()));
+        newDimension.convention(ImmutableSet.copyOf(WindowsDistribution.values()));
+        ((ExtensionAware) library).getExtensions().add("targetWindowsDistributions", newDimension);
```
lacasseio, line 121: 🔍 See conversation in the `#ext-nokee` Slack channel around `ExtensionAware` guideline.

lacasseio:
@big-guy I released an integration version. We still need to figure out the dependency verification error for CI to be happy.

buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
lptr: We have some special tests where `-Xcheck:jni` output is treated as a reason for failure. These are the tests marked with `JniChecksEnabled`. For other tests that were introduced earlier, we did not enforce these checks yet, and they do fail.

One option to remove this distinction would be to make sure none of the current code emits any JNI warnings.

buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
```diff
@@ -0,0 +1,213 @@
...
+        VariantsExtension variants = project.getExtensions().getByType(VariantsExtension.class);
+        variants.getVariantNames().set(library.getVariants().flatMap(it -> {
+            // Only depend on variants which can be built on the current machine
+            boolean onlyLocalVariants = project.getProviders().gradleProperty("onlyLocalVariants").forUseAtConfigurationTime().isPresent();
+            if (onlyLocalVariants && !it.getSharedLibrary().isBuildable()) {
+                return ImmutableList.of();
+            } else {
+                return ImmutableList.of(toVariantName(it.getTargetMachine()));
+            }
+        }));
```
big-guy, line 66: What does it mean to set the "variant names" here? This disables certain variants from running?

native-platform/build.gradle
```diff
@@ -20,41 +23,15 @@ javadoc {
...
-                }
-            }
+library {
+    baseName = 'native-platform'
+    sources.configureEach(CppSourceSet) {
+        source.from('src/shared/cpp', 'src/main/cpp')
+        headers.from('src/shared/headers', 'build/generated/jni-headers/main')
+    }
+    variants.configureEach {
+        if (NcursesVersion.values().any { buildVariant.hasAxisOf(it) }) {
```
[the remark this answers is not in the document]
lacasseio: It's a little bit out of date but it should give you a good enough idea: https://docs.nokee.dev/dsl/dev.nokee.platform.jni.JavaNativeInterfaceLibrary.html

PR #299, opened 2021-12-10
Expose functionality to stop watching moved paths on Linux

lptr:
When the parent of a watched directory is moved to a different location on Linux, the kernel does not report any event. This is a similar problem to what we had on Windows: https://github.com/gradle/native-platform/pull/290.

The solution is similar, too: add a `LinuxFileWatcher.stopWatchingMovedPaths()` method.

### DIFF
--- a/file-events/src/test/groovy/net/rubygrapefruit/platform/file/MovedDirectoriesFileEventFunctionsTest.groovy
+++ b/file-events/src/test/groovy/net/rubygrapefruit/platform/file/MovedDirectoriesFileEventFunctionsTest.groovy
@@ -25,6 +25,7 @@ import java.nio.file.Files
 import java.util.regex.Pattern
 
 import static java.util.logging.Level.WARNING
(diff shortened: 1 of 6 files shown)


file-events/src/test/groovy/net/rubygrapefruit/platform/file/MovedDirectoriesFileEventFunctionsTest.groovy
```diff
@@ -0,0 +1,160 @@
...
+        expectNoEvents()
+
+        when:
+        droppedPaths = watcher.stopWatchingMovedPaths()
+        then:
+        droppedPaths == []
+    }
+
+    @Requires({ Platform.current().linux || Platform.current().windows })
+    def "stops watching when parent of path is moved"() {
```
wolfs, line 133: Can you add a test here that on macOs, when we move a parent of the watched directory we receive file events for the watched directory? And when we then create a new directory at the location of the watched directory we receive events?

wolfs (commented):
LGTM! I left one comment about the test coverage.

PR #300, opened 2021-12-13
Allow checking for moved paths on Linux

wolfs:
A variation on #299, which allows passing in a list of locations to check. This avoids checking all the watched paths, which may be quite expensive on Linux.

### DIFF
--- a/file-events/src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
+++ b/file-events/src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
@@ -372,23 +372,6 @@ class BasicFileEventFunctionsTest extends AbstractFileEventFunctionsTest {
         expectEvents change(CREATED, targetFileInside)
     }
 
-    @Issue("https://github.com/gradle/native-platform/issues/193")
-    def "can rename watched directory"() {
-        given:
-        def watchedDirectory = new File(rootDir, "watched")
-        watchedDirectory.mkdirs()
-        startWatcher(watchedDirectory)
-
-        when:
-        watchedDirectory.renameTo(new File(rootDir, "newWatched"))
-        waitForChangeEventLatency()
-        then:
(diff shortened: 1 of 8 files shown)


file-events/src/test/groovy/net/rubygrapefruit/platform/file/BasicFileEventFunctionsTest.groovy
```diff
@@ -417,15 +415,18 @@ class BasicFileEventFunctionsTest extends AbstractFileEventFunctionsTest {
     def "fails when watching non-existent directory"() {
         given:
         def missingDirectory = new File(rootDir, "missing")
+        def message = Platform.current().linux
+            ? "Couldn't stat path"
+            : "Couldn't add watch"
```
lptr, line 420: I wonder if we should standardize the messages themselves instead.

wolfs: Yes, I think it would be better to standardize here. The message should probably be `Couldn't add watch`, since that what we tried to do. The user of the API doesn't know why we needed to stat the thing we wanted to watch.

lptr: Agreed, though it would make sense to somehow also capture that the error message belongs to `stat()` (as opposed to `inotify_add_watch()`) for disambiguation later. So something like:

- `Couldn't add watch (stat() failed with errno = 2)`
- `Couldn't add watch (inotify_add_watch() failed with errno = 3)`

WDYT?

wolfs: Sounds good. I'll make the change.

wolfs: Done.

lptr (commented):
LGTM.

PR #301, opened 2021-12-15
Report unwatched directories as moved on stopWatchingMovedPaths

wolfs:
If a directory isn't watched any more on Linux, it should be reported by `stopWatchingMovedPaths()`, so the client can drop he state it retains for the unwatched directory as well.

### DIFF
--- a/file-events/src/file-events/cpp/linux_fsnotifier.cpp
+++ b/file-events/src/file-events/cpp/linux_fsnotifier.cpp
@@ -387,10 +387,10 @@ Java_net_rubygrapefruit_platform_internal_jni_LinuxFileEventFunctions_isGlibc0(J
 }
 
 JNIEXPORT void JNICALL
-Java_net_rubygrapefruit_platform_internal_jni_LinuxFileEventFunctions_00024LinuxFileWatcher_stopWatchingMovedPaths0(JNIEnv* env, jobject, jobject javaServer, jobjectArray absolutePathsToCheck, jobject jDroppedPaths) {
+Java_net_rubygrapefruit_platform_internal_jni_LinuxFileEventFunctions_00024LinuxFileWatcher_stopWatchingMovedPaths0(JNIEnv* env, jobject, jobject javaServer, jobjectArray jAbsolutePathsToCheck, jobject jDroppedPaths) {
     try {
         Server* server = (Server*) getServer(env, javaServer);
-        server->stopWatchingMovedPaths(absolutePathsToCheck, jDroppedPaths);
+        server->stopWatchingMovedPaths(jAbsolutePathsToCheck, jDroppedPaths);
     } catch (const exception& e) {
         rethrowAsJavaException(env, e);
     }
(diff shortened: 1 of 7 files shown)


file-events/src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -348,15 +354,17 @@ void Server::stopWatchingMovedPaths(const vector<u16string>& absolutePathsToChec
...
-        getJavaExceptionAndPrintStacktrace(env);
+        addToList(env, droppedPaths, jPathToCheck);
 
         watchPoint.cancel();
     }
 }
 
+void Server::addToList(JNIEnv* env, jobject jList, jstring jString) {
+        env->CallBooleanMethod(jList, listAddMethod, jString);
+        getJavaExceptionAndPrintStacktrace(env);
```
lptr, line 365: Why do we get exception and print stacktrace here?

file-events/src/file-events/cpp/linux_fsnotifier.cpp
```diff
@@ -380,11 +388,9 @@ Java_net_rubygrapefruit_platform_internal_jni_LinuxFileEventFunctions_isGlibc0(J
 }
 
 JNIEXPORT void JNICALL
-Java_net_rubygrapefruit_platform_internal_jni_LinuxFileEventFunctions_00024LinuxFileWatcher_stopWatchingMovedPaths0(JNIEnv* env, jobject, jobject javaServer, jobjectArray jAbsolutePathsToCheck, jobject jDroppedPaths) {
+Java_net_rubygrapefruit_platform_internal_jni_LinuxFileEventFunctions_00024LinuxFileWatcher_stopWatchingMovedPaths0(JNIEnv* env, jobject, jobject javaServer, jobjectArray absolutePathsToCheck, jobject jDroppedPaths) {
```
lptr, line 391: Why remove the `j` prefix?

file-events/src/file-events/cpp/linux_fsnotifier.cpp
wolfs: I suppose we don't want to fail for some reason. This was like that before. Do you know why?

lptr: Ah, I copied it over from somewhere else. We should remove this.

lptr: I think we just remove this call. A Java exception is just a thread-local, basically. It will be passed to the unexpected exception handler associated with the thread if one occurs when we return from the JNI call anyway.

file-events/src/file-events/cpp/linux_fsnotifier.cpp
wolfs: No particular reason. I re-added the `j`.

file-events/src/file-events/cpp/linux_fsnotifier.cpp
wolfs: Ok, I removed the logging of the stacktrace.

wolfs: Aha, when I remove this I get:

```
WARNING in native method: JNI call made without checking exceptions when required to from CallBooleanMethodV
```

What is the right approach here?

wolfs: I think now I got it right: I am throwing now `JavaExceptionThrownException` and catch it in the JNI method to have the right control flow.

wolfs: I moved back to the old handling.

wolfs: We talked again, and since we need to do something with the exception because of the JNI checks, we decided that throwing it is the right thing to do.

PR #304, opened 2022-02-08
Completes the migration to Nokee plugins

lacasseio:
This PR relies on a locally published Nokee distribution with two minor fixes: 1) predicate filter for build variant and 2) ignores dimension with empty names when computing names. The final solution for those fixes will soon be integrated into master.

This PR focus on migrating the support for FreeBSD, Ncurses, publications and testJni task. All build features haven't been tested yet which is the next step in the migration.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -10,7 +10,7 @@ repositories {
 
 dependencies {
     implementation gradleApi()
-    implementation platform('dev.nokee:nokee-gradle-plugins:0.4.1229-202112011142.4a01703c')
+    implementation platform('dev.nokee:nokee-gradle-plugins:0.4.1509-202202081144.9cd514f1')
     implementation 'org.apache.httpcomponents.client5:httpclient5:5.0.1'
     implementation 'com.google.guava:guava:28.2-jre'
     implementation 'org.gradle:test-retry-gradle-plugin:1.1.8'
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -52,5 +52,9 @@ gradlePlugin {
             id = "gradlebuild.ncurses"
             implementationClass = "gradlebuild.NcursesPlugin"
         }
+        ncursesRuntime {
+            id = "gradlebuild.ncurses-runtime"
+            implementationClass = "gradlebuild.NcursesRuntimePlugin"
+        }
     }
 }
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -14,6 +14,7 @@ dependencies {
     implementation 'org.apache.httpcomponents.client5:httpclient5:5.0.1'
     implementation 'com.google.guava:guava:28.2-jre'
     implementation 'org.gradle:test-retry-gradle-plugin:1.1.8'
+    implementation 'org.apache.commons:commons-lang3:3.12.0'
 }
 
 java {
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -10,7 +10,7 @@ repositories {
 
 dependencies {
     implementation gradleApi()
-    implementation platform('dev.nokee:nokee-gradle-plugins:0.4.1560-202202152015.2a7d9703')
+    implementation platform('dev.nokee:nokee-gradle-plugins:0.4.1561-202202152051.db45d178')
     implementation 'org.apache.httpcomponents.client5:httpclient5:5.0.1'
     implementation 'com.google.guava:guava:28.2-jre'
     implementation 'org.gradle:test-retry-gradle-plugin:1.1.8'
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -44,6 +44,10 @@ gradlePlugin {
             id = "gradlebuild.freebsd"
             implementationClass = "gradlebuild.FreeBsdPlugin"
         }
+        freebsdRuntime {
+            id = "gradlebuild.freebsd-runtime"
+            implementationClass = "gradlebuild.FreeBsdRuntimePlugin"
+        }
         ncurses {
             id = "gradlebuild.ncurses"
             implementationClass = "gradlebuild.NcursesPlugin"
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -10,7 +10,7 @@ repositories {
 
 dependencies {
     implementation gradleApi()
-    implementation platform('dev.nokee:nokee-gradle-plugins:0.4.1509-202202081144.9cd514f1')
+    implementation platform('dev.nokee:nokee-gradle-plugins:0.4.1560-202202152015.2a7d9703')
     implementation 'org.apache.httpcomponents.client5:httpclient5:5.0.1'
     implementation 'com.google.guava:guava:28.2-jre'
     implementation 'org.gradle:test-retry-gradle-plugin:1.1.8'
--- a/buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
+++ b/buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
@@ -47,6 +47,7 @@ import org.gradle.nativeplatform.toolchain.NativeToolChainRegistry;
 import org.gradle.nativeplatform.toolchain.VisualCpp;
 
 import java.io.File;
+import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.Map;
 import java.util.Set;
 
@@ -215,10 +216,11 @@ public abstract class JniNokeePlugin implements Plugin<Project> {
         // We register the publications here, so they are available when the project is used as a composite build.
         final Provider<String> artifactId = project.getTasks().named("jar", Jar.class)
             .flatMap(Jar::getArchiveBaseName);
-        final Provider<Iterable<Jar>> publishableJarTasks = library(project).flatMap(allVariants())
+        final ListProperty<Jar> publishableJarTasks = project.getObjects().listProperty(Jar.class).value(library(project).flatMap(allVariants())
             .map(toPublishableVariantGroups())
-            .map(toBuildableVariantUsingVariantNames(variants.getVariantNames()))
-            .flatMap(toPublishableJars(project, artifactId));
+            .map(toBuildableVariantUsingBuildableExtension())
+            .flatMap(toPublishableJars(project, artifactId)));
+        publishableJarTasks.finalizeValueOnRead(); // to minimize computation of provided value
 
         project.getConfigurations().named("runtimeElements", configuration -> {
             final Provider<? extends Iterable<PublishArtifact>> publishableJars = project.getObjects().listProperty(PublishArtifact.class).value(publishableJarTasks.map(it -> stream(it).map(ArchivePublishArtifact::new).collect(toList())));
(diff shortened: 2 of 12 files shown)


buildSrc/build.gradle
```diff
@@ -10,10 +10,11 @@ repositories {
 
 dependencies {
     implementation gradleApi()
-    implementation platform('dev.nokee:nokee-gradle-plugins:0.4.1229-202112011142.4a01703c')
+    implementation platform('dev.nokee:nokee-gradle-plugins:0.4.1509-202202081144.9cd514f1')
     implementation 'org.apache.httpcomponents.client5:httpclient5:5.0.1'
     implementation 'com.google.guava:guava:28.2-jre'
     implementation 'org.gradle:test-retry-gradle-plugin:1.1.8'
+    implementation 'org.apache.commons:commons-lang3:3.12.0'
```
lacasseio, line 17: 🔍 I use `SystemUtils` instead of using internal Gradle classes.

buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
```diff
@@ -45,73 +70,111 @@ public void apply(Project project) {
...
+                            return true; // no known ncurses versions, assuming buildable
+                        } else {
+                            // For each variant with the ncurses dimension, check if the version is available
+                            if (variant.getBuildVariant().hasAxisOf(NCURSES_5)) {
+                                return availableNcursesVersions.get().contains(NCURSES_5);
+                            } else if (variant.getBuildVariant().hasAxisOf(NCURSES_6)) {
+                                return availableNcursesVersions.get().contains(NCURSES_6);
+                            }
+                            return true;
+                        }
```
lacasseio, line 125: 🔍 Nokee doesn't support the ad-hoc buildable flag, instead we need to verify the buildability manually for custom dimensions. In the future, we will move toward including the compile/link dependencies to verify the buildability. If the dependencies do not resolve successfully, we can assert the variant is not buildable. It opens the unfortunate possibility that users force dependencies to resolves too early, however, early resolution of provider check should take care of this issue.

buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
```diff
@@ -175,6 +225,158 @@ private void configureNativeVersionGeneration(Project project) {
...
+            .flatMap(Jar::getArchiveBaseName);
+        final Provider<Iterable<Jar>> publishableJarTasks = library(project).flatMap(allVariants())
+            .map(toPublishableVariantGroups())
+            .map(toBuildableVariantUsingVariantNames(variants.getVariantNames()))
+            .flatMap(toPublishableJars(project, artifactId));
+
+        project.getConfigurations().named("runtimeElements", configuration -> {
+            final Provider<? extends Iterable<PublishArtifact>> publishableJars = project.getObjects().listProperty(PublishArtifact.class).value(publishableJarTasks.map(it -> stream(it).map(ArchivePublishArtifact::new).collect(toList())));
+            configuration.getOutgoing().getArtifacts().addAllLater(publishableJars);
+        });
```
lacasseio, line 241: 🔍 This interesting side effect of focusing on provider mapping instead of using `afterEvaluate`. Despite the interesting side effect, Gradle still lack some fundamentals to properly use `addAllLater` to effectively defer everything until needed. In this case, we use the internal `ArchivePublishArtifact` type to show how far we could go. Unfortunately, for the publications below, it's not possible.

buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
```diff
@@ -175,6 +225,158 @@ private void configureNativeVersionGeneration(Project project) {
...
+            for (Jar nativeJar : publishableJarTasks.get()) {
+                final String variantName = nativeJar.getName().substring("jar-".length());
+                project.getExtensions().configure(PublishingExtension.class, publishing -> publishing.publications(publications -> publications.create(variantName, MavenPublication.class, publication -> {
+                    publication.artifact(nativeJar);
+                    publication.artifact(emptyZip.get(), it -> it.setClassifier("sources"));
+                    publication.artifact(emptyZip.get(), it -> it.setClassifier("javadoc"));
+                    publication.setArtifactId(nativeJar.getArchiveBaseName().get());
+                })));
+            }
+        });
```
lacasseio, line 253: 🔍 We could eliminate the `afterEvaluate` if we could successfully use `addAllLater` and map our `publishableTasks` provider into a list of publications. The alternative that we experimented with in Nokee is a lazy stream-like interface that allows to either collect into a `Provider` or iterate as new values come into existence. It would be very similar and would look something like the following:

```
Stream<Jar> publishableJarTaks = ...
publishableJarTasks.forEach { nativeJar ->
    project.publishing.publications.create('<name>', MavenPublication) {
        artifact(nativeJar)
        //...
    }
}
```

lacasseio: The streaming concept is something we are looking into adding to Nokee. We unfortunately don't have an ETA.

buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
```diff
@@ -202,10 +404,7 @@ void createToolChains(NativeToolChainRegistry toolChainRegistry) {
             });
             toolChainRegistry.create("clang", Clang.class, toolChain -> {
                 // The core Gradle toolchain for Clang only targets x86 and x86_64 out of the box.
-                OperatingSystem os = new DefaultNativePlatform("current").getOperatingSystem();
-                if (os.isMacOsX()) {
-                    toolChain.target("macosaarch64");
-                }
+                toolChain.target("macosaarch64");
```
lacasseio, line 407: ❌ There is still some work to do here to avoid building aarch64 on x64 architecture.

buildSrc/src/main/java/gradlebuild/NcursesRuntimePlugin.java
```diff
@@ -0,0 +1,109 @@
...
+    private SetProperty<NcursesVersion> registerNcursesDimension(JavaNativeInterfaceLibrary library) {
+        return library.getDimensions().newAxis(NcursesVersion.class, it -> it.filter(t -> {
+            if (t.hasAxisOf(library.getMachines().getWindows().getOperatingSystemFamily())) {
+                return t.hasAxisOf(NCURSES_5) || t.hasAxisOf(NCURSES_6);
+            } else if (t.hasAxisOf(library.getMachines().getLinux().getOperatingSystemFamily())) {
+                return false;
+            } else {
+                return t.hasAxisOf(NCURSES_6);
+            }
+        }));
```
lacasseio, line 78: ❌ The filter here will change once we properly implement the feature in Nokee.

buildSrc/src/main/java/gradlebuild/NcursesRuntimePlugin.java
```diff
@@ -0,0 +1,109 @@
...
+        project.getPluginManager().withPlugin("dev.nokee.jni-library", appliedPlugin -> {
+            library(project, library -> {
+                // Register ncurses dimension
+                SetProperty<NcursesVersion> ncurses = registerNcursesDimension(library);
+
+                // Defaults to all ncurses versions
+                ncurses.convention(ImmutableSet.copyOf(NcursesVersion.values()));
+
+                // Register dimension as library extensions
+                ((ExtensionAware) library).getExtensions().add("targetNcurses", ncurses);
```
lacasseio, line 43: ❓ I'm still waiting for guidance on the usage for `ExtensionAware` by plugin authors. See https://gradle.slack.com/archives/C02FZQ5EZDE/p1638225105038300

buildSrc/src/main/java/gradlebuild/NcursesRuntimePlugin.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private static Callable<Iterable<String>> ofNcursesLibraryFlags(TargetMachineFactory machines, BuildVariant buildVariant) {
+        return () -> {
+            if (buildVariant.hasAxisOf(machines.getLinux().getOperatingSystemFamily()) && !buildVariant.hasAxisOf(NCURSES_5)) {
+                return ImmutableList.of("-lncursesw");
+            } else {
+                return ImmutableList.of("-lcurses");
+            }
+        };
```
lacasseio, line 66: 🔍 In theory, `curses`/`ncurses` library should be a normal dependency (added via `dependencies` DSL block). The main reason is allowing Nokee to use the dependency information to determine buildability among other things. This work will be done in the future with a deprecation cycle.

buildSrc/src/main/java/gradlebuild/actions/MixInJavaNativeInterfaceLibraryProperties.java
```diff
@@ -0,0 +1,45 @@
...
+
+        library.getSources().configureEach(CppSourceSet.class, sourceSet -> {
+            sourceSet.from(cppSources);
+            ((HasConfigurableHeaders) sourceSet).getHeaders().from(privateHeaders);
+
+            // We rewire the generated JNI headers here
+            //   because the convention gets overwritten on the first {@literal from} call.
+            //   The convention idea is something that will be phased out in the near future.
+            ((HasConfigurableHeaders) sourceSet).getHeaders().from(project.getTasks().named("compileJava", JavaCompile.class).flatMap(it -> it.getOptions().getHeaderOutputDirectory()));
+        });
```
lacasseio, line 43: 🔍 Soon Nokee will allow dynamic mix-in which the language plugins will use. The mix-in behaviour for this particular case will look very similar to the code here. At that point, we will be able to remove this code.

native-platform/build.gradle
```diff
@@ -25,13 +24,14 @@ javadoc {
...
-    sources.configureEach(CppSourceSet) {
-        source.from('src/shared/cpp', 'src/main/cpp')
-        headers.from('src/shared/headers')
-    }
+    cppSources.setFrom('src/shared/cpp', 'src/main/cpp')
+    privateHeaders.setFrom('src/shared/headers')
     variants.configureEach {
         if (NcursesVersion.values().any { buildVariant.hasAxisOf(it) }) {
             baseName = 'native-platform-curses'
+            sources.configureEach(dev.nokee.language.cpp.CppSourceSet) {
```
lacasseio, line 32: ❓ It would be important to discuss how plugin authors are expected to create default imports. We experimented with various solutions but they range in complexity and efficiency. A core feature would be much better.

buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
```diff
@@ -163,6 +209,10 @@ private void configureNativeVersionGeneration(Project project) {
             task.getGeneratedJavaSourcesDir().set(new File(generatedFilesDir, "version/java"));
             task.getVersionClassPackageName().set(nativeVersion.getVersionClassPackageName());
             task.getVersionClassName().set(nativeVersion.getVersionClassName());
+            task.getNativeSources().from(
+                library(project).map(JavaNativeInterfaceLibraryUtils::cppSources),
+                library(project).map(JavaNativeInterfaceLibraryUtils::privateHeaders)
+            );
```
lacasseio, line 215: ❓ The native sources no longer include the auto-generated JNI headers as they were causing a cycle in the task graph. In theory, the JNI headers are implicitly covered by the C++ sources given they implement the prototype hence have the same signature. It's up to you to validate if this change is correct or not.

lacasseio (commented):
I left some additional information on the changes.

PR #305, opened 2022-02-11
Test file events on xfs and btrfs

wolfs:
I added an xfs and a btrfs partition to the Ubuntu amd64 AMIs. We can now run the file event tests there as well to see that those two file systems are supported as well.

### DIFF
--- a/.teamcity/extensions.kt
+++ b/.teamcity/extensions.kt
@@ -97,6 +97,12 @@ val Agent.allPublishTasks
         else -> publishJniTasks
     }
 
+val Agent.extraTestTasks
+    get() = when (this) {
+        Agent.UbuntuAmd64 -> " :file-events:testBtrfs :file-Events:testXfs"
+        else -> ""
+    }
+
 fun BuildFeatures.lowerRequiredFreeDiskSpace() {
     freeDiskSpace {
         // Configure less than the default 3GB, since the disk of the agents is only 5GB big.
--- a/.teamcity/extensions.kt
+++ b/.teamcity/extensions.kt
@@ -99,7 +99,7 @@ val Agent.allPublishTasks
 
 val Agent.extraTestTasks
     get() = when (this) {
-        Agent.UbuntuAmd64 -> " :file-events:testBtrfs :file-Events:testXfs"
+        Agent.UbuntuAmd64 -> " :file-events:testBtrfs :file-events:testXfs"
         else -> ""
     }
 
--- a/buildSrc/src/main/java/gradlebuild/NativePlatformComponentPlugin.java
+++ b/buildSrc/src/main/java/gradlebuild/NativePlatformComponentPlugin.java
@@ -18,8 +18,12 @@ import org.gradle.api.tasks.testing.Test;
 import org.gradle.testretry.TestRetryPlugin;
 import org.gradle.testretry.TestRetryTaskExtension;
 
+import java.io.File;
+import java.util.stream.Stream;
+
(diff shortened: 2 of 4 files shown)


asodja (approved):
LGTM

.teamcity/extensions.kt
```diff
@@ -97,6 +97,12 @@ val Agent.allPublishTasks
         else -> publishJniTasks
     }
 
+val Agent.extraTestTasks
+    get() = when (this) {
+        Agent.UbuntuAmd64 -> " :file-events:testBtrfs :file-Events:testXfs"
```
lptr, line 102:
```suggestion
        Agent.UbuntuAmd64 -> " :file-events:testBtrfs :file-events:testXfs"
```

lptr:
How does this work? Where is the code that actually makes those locations be xfs or btrfs?

wolfs:
> How does this work? Where is the code that actually makes those locations be xfs or btrfs?

When we bake the Ubuntu AMI, we now add two partitions to the SSD, formatted with XFS and btrfs:
https://github.com/gradle/dev-infrastructure/blob/5f9e66ea0eafb1ffc648f92646f881e1cbfc2b51/salt/ami-baker/common/ssd-startup.sh

lptr:
Thanks. I think this is pretty easy to break, though. If for whatever reason those mounts change, we'll just be running these tests on extfs without noticing. I think it would be good to add some check, even just to parse the output of `mount` or something.

We could do the same for `SUBST` on Windows.

buildSrc/src/main/java/gradlebuild/NativePlatformComponentPlugin.java
```diff
@@ -67,11 +71,24 @@ private void configureTestTasks(Project project) {
             Provider<String> agentName = project.getProviders().gradleProperty("agentName").forUseAtConfigurationTime();
             test.systemProperty("agentName", agentName.getOrElse("Unknown"));
         });
+
+        Stream.of("xfs", "btrfs").forEach(fileSystem -> {
+            project.getTasks().register("test" + capitalize(fileSystem), Test.class, test -> {
+                test.systemProperty(TEST_DIRECTORY_SYSTEM_PROPERTY, new File("/" + fileSystem, "native-platform/test files").getAbsolutePath());
+            });
```
lptr, line 78: Can we check the file system type here? We can do something simple, like check the output of `findmnt -n -o FSTYPE -T /${fileSystem}`

There are more options on how to determine the file system from the path here: https://unix.stackexchange.com/questions/21799/how-can-i-determine-the-fs-type-of-my-current-working-directory

lptr:
Sorry, I mixed up `SUBST` and our FAT checks. For those we could add a similar test, no?

buildSrc/src/main/java/gradlebuild/NativePlatformComponentPlugin.java
lptr: The reason I think we should check here is that the test is far away from here. For me checking if the infrastructure is what we expect belongs to the build, not so much to a specific test that is supposed to verify if the code itself is correct.

If we want to keep the test as a unit test, I think it should be its own class, and have comments linking back to here explaining why this check is required.

wolfs:
> Sorry, I mixed up `SUBST` and our `FAT` checks. For those we could add a similar test, no?

We could, though those tests are in `gradle/gradle`. We currently don't run tests against FAT in native platform - though maybe we should...

lptr (approved):
LGTM.

PR #306, opened 2022-02-16
Disable Linux AARCH64 target on other targets

lacasseio:


buildSrc/src/main/java/gradlebuild/JniNokeePlugin.java
```diff
@@ -395,7 +395,11 @@ void createToolChains(NativeToolChainRegistry toolChainRegistry) {
             toolChainRegistry.create("gcc", Gcc.class, toolChain -> {
                 // The core Gradle toolchain for gcc only targets x86 and x86_64 out of the box.
                 // https://github.com/gradle/gradle/blob/36614ee523e5906ddfa1fed9a5dc00a5addac1b0/subprojects/platform-native/src/main/java/org/gradle/nativeplatform/toolchain/internal/gcc/AbstractGccCompatibleToolChain.java
-                toolChain.target("linuxaarch64");
+                toolChain.target("linuxaarch64", platformToolChain -> {
```
wolfs: Do you also need to change the amd64 to not build when on aarch64?

lacasseio: I'm a bit unsure about that. In theory, we should however I don't know if Gradle has any special handling in the toolchain detection. At the moment, the failures seem to be pointing at something else than this configuration so it's unclear. Definitely something to keep in mind.

big-guy:
@wolfs did something change with DCO?

PR #311, opened 2022-03-02
Update to 6.9.1

big-guy:


buildSrc/src/main/java/gradlebuild/JniPlugin.java
```diff
@@ -287,6 +288,12 @@ void addComponentSourcesSetsToProjectSourceSet(ModelMap<Task> tasks, ModelMap<So
         @Mutate void configureBinaries(@Each NativeBinarySpecInternal binarySpec) {
             DefaultNativePlatform currentPlatform = new DefaultNativePlatform("current");
             Architecture currentArch = currentPlatform.getArchitecture();
+            // TODO: There's a mismatch between the names of 64-bit ARM architectures between
+            // native-platform and the native plugins in Gradle. If we're using the Gradle name,
+            // switch it to the native-platform version
+            if (currentArch.getName().equals("arm-v8")) {
+                currentArch = new org.gradle.nativeplatform.platform.internal.DefaultArchitecture("aarch64");
```
wolfs, line 295: Does this mean something is broken in Gradle/native platform when returning the architecture for M1 macs? Especially given that the JDK returns `aarch64`?

big-guy: Yes, maybe not "broken", but I think our names are incomplete. 
I added the aarch64 alias in https://github.com/gradle/gradle/pull/20086

and tried reverting this change: https://builds.gradle.org/viewQueued.html?itemId=49919245

PR #314, opened 2022-07-07
Adjust HardLinkFileEventFunctionsTest for macOS 12

wolfs:
It looks like when you change files which are hardlinks on macOS 12, then you get events for all the paths which are hard linked to the same location, as opposed to only for the path you changed. That is good.

What is not so good is that it seems in the situation where the path you changed is outside the watched directory, and a path hard linked to the file inside the watched directory, then you receive an event for the file outside of the watched directory but with the wrong path.

E.g. in this situation:
```

\ - link
\ - watched
   \ - target
```
When you watch `watched` and you change `target`, then you get two events: one for `watched/target` (good) and another one for `watched/link`, a file which doesn't even exist.

This isn't very bad for our code, since we only drop more things, though it also isn't good.


file-events/src/test/groovy/net/rubygrapefruit/platform/file/HardLinkFileEventFunctionsTest.groovy
```diff
@@ -108,8 +108,9 @@ class HardLinkFileEventFunctionsTest extends AbstractFileEventFunctionsTest {
         then:
         // On Windows we seem to be getting multiple MODIFIED events
         // On Linux we _sometimes_ seem to be getting multiple events
+        // For macOS 12 we get events for both hardlink locations
```
lptr, line 111: I wonder how/when we should check back here if this gets fixed in macOS 12.x or 13+...

PR #318, opened 2022-08-31
Add support for running on ARM64 Windows

modmuss50:
Hello 👋,

This PR is a big step forward in supporting ARM64 Windows.

Depends on: https://github.com/gradle/gradle/pull/21780
Major step forward for: https://github.com/gradle/gradle/issues/21703

Screenshot of the native-platform-test application running:
image

This will require the ARM64 compiler to be installed using the Visual Studio installer, I find this is not usually included by default. This has currently only been tested using the test application, and not "end-to-end" in a full Gradle project.

I look forward to any questions or feedback 👍 

Signed-off-by: modmuss50 <modmuss50@gmail.com>

### DIFF
--- a/file-events/src/file-events/cpp/file-events-version.cpp
+++ b/file-events/src/file-events/cpp/file-events-version.cpp
@@ -1,7 +1,7 @@
 #include "native_platform_version.h"
-#include "net_rubygrapefruit_platform_internal_jni_AbstractFileEventFunctions.h"
+#include "net_rubygrapefruit_platform_internal_jni_AbstractNativeFileEventFunctions.h"
 
 JNIEXPORT jstring JNICALL
-Java_net_rubygrapefruit_platform_internal_jni_AbstractFileEventFunctions_getVersion0(JNIEnv* env, jclass) {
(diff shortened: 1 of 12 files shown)


modmuss50:
This PR should be ready for review now that https://github.com/gradle/gradle/pull/21780 was merged, however I think a maintainer needs to finish it off.

- native-platform needs updating to Gradle 8 and the latest nightly, I'm struggling with this.
- The build machines will need the Visual Studio ARM64 components installed.
- After merging native-platform needs updating in Gradle.

This will fix the `WindowsRegistry is not supported on this operating system` when running on ARM64.

imkiva:
Hi maintainers, is there any plan for merging this PR? I met this problem today when I was trying to setup my Windows on ARM developer kit. Or is that possible to bypass this restriction through some configurable settings in `gradle.properties`? I know one use case of windows registry is to detect installed JDKs but I am not sure if it is used elsewhere.

imkiva:
Maybe it is necessary to upgrade the Gradle wrapper version to 7.6.1 or higher since it contains the https://github.com/gradle/gradle/pull/21780 needed by this patch.

Issue #319, opened 2022-10-20
When native file-event run loop exits abnormally events appear incorrectly on the event queue


lptr:
When the run loop exits normally, it should publish a `TerminateEvent` "poison pill" to the event queue. The consumer of the events knows when receiving that event to stop processing more events.

When a native exception happens, causing the run loop to exit, we publish a `FailureEvent` that should tell the consumer that some kind of error happened, and act accordingly.

However, due to a bug, when an exception happens, we actually publish a `TerminateEvent` before the `FailureEvent`, causing the consumer to digest the "poison pill", stop receiving events, and thus will never see the `FailureEvent`.

PR #320, opened 2022-10-20
Report run loop termination errors before reporting the termination itself

lptr:
Publish `TerminateEvent` after any errors related to stopping the native processor thread have been reported to the queue. This prevents the thread processing the event queue from exiting before seeing the `FailureEvent`.

Fixes https://github.com/gradle/native-platform/issues/319.


wolfs (approved):
LGTM!

PR #321, opened 2022-10-21
Test file event handling termination

lptr:
In https://github.com/gradle/native-platform/pull/320 we've moved the reporting of the file event handling run loop's termination to the Java side. This PR separates the non-native parts of `AbstractFileEventFunctions` and `AbstractFileWatcher` and adds synthetic tests without a native backend to check how the Java side of event handling works on its own.


wolfs (approved):
LGTM!

PR #331, opened 2023-09-06
Add a flag for PosixFiles.getMode to follow symlinks

ov7a:


native-platform/src/test/groovy/net/rubygrapefruit/platform/file/PosixFilesTest.groovy
```diff
@@ -317,6 +317,29 @@ class PosixFilesTest extends FilesTest {
         e.message == "Could not read symlink $symlinkFile: could not readlink (errno 22: Invalid argument)"
     }
 
+    @Unroll
+    def "can get mode for a file behind symbolic link"() {
```
lptr, line 321: :x: We should add a test with a broken symlink, too. We should test both the new method and the old method.

native-platform/src/main/java/net/rubygrapefruit/platform/internal/DefaultPosixFiles.java
```diff
@@ -66,8 +66,12 @@ public void setMode(File file, int perms) {
         }
     }
 
-    public int getMode(File file) {
-        PosixFileInfo stat = stat(file);
+    public int getMode(File file) throws NativeException {
+        return getMode(file, false);
+    }
+
+    public int getMode(File file, Boolean linkTarget) throws NativeException {
```
lptr, line 73: :x: why is this `Boolean` instead of `boolean`?

ov7a: Nice catch, that's a typo

native-platform/src/test/groovy/net/rubygrapefruit/platform/file/PosixFilesTest.groovy
ov7a: I'll add a test for a broken link, ok.

Both implementations are covered by L333-337, which case do you miss

lptr: They cover it for an non-borken link. We should cover the both old and the new method for a broken link as well.

ov7a: I've added a test.

native-platform/src/test/groovy/net/rubygrapefruit/platform/file/PosixFilesTest.groovy
```diff
@@ -317,6 +317,53 @@ class PosixFilesTest extends FilesTest {
...
+        files.getMode(brokenSymlinkFile)
+
+        then:
+        noExceptionThrown()
+
+        when:
+        files.getMode(brokenSymlinkFile, true)
+
+        then:
+        noExceptionThrown()
```
lptr, line 360: :x: This doesn't make sense to me. The caller asks to follow the symlink explicitly, but it is not possible to follow the symlink, because it's broken. This should throw an exception, just as if you asked for a non-existent file in the first place, not silently return the permissions for the referring symlink.

ov7a:
FTR, file-events tests are failing but and don't see any connection this my

native-platform/src/main/java/net/rubygrapefruit/platform/file/PosixFiles.java
```diff
@@ -43,6 +43,15 @@ public interface PosixFiles extends Files, NativeIntegration {
     @ThreadSafe
     int getMode(File path) throws NativeException;
 
+    /**
+     * Gets the mode for the given file.
+     *
+     * @param linkTarget When true and the file is a symlink, return details of the target of the symlink instead of details of the symlink itself.
+     * @throws NativeException On failure.
```
lptr, line 50: 💅 I think we could be more specific in when we throw the exception here.

lptr (approved):
LGTM.

PR #332, opened 2023-09-08
Remove `pgpSigningKey` param in Kotlin DSL code

blindpirate:
Let's move it to `Gradle Native` TeamCity project level (https://builds.gradle.org/admin/editProject.html?projectId=GradleNative&tab=projectParams), so we can update the param without touching the code.

### DIFF
--- a/.teamcity/NativePlatformPublishing.kt
+++ b/.teamcity/NativePlatformPublishing.kt
@@ -61,8 +61,6 @@ open class NativePlatformPublishSnapshot(releaseType: ReleaseType, uploadTasks: 
     params {
         param("ARTIFACTORY_USERNAME", releaseType.username)
         password("ARTIFACTORY_PASSWORD", releaseType.password, display = ParameterDisplay.HIDDEN)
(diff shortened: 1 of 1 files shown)


.teamcity/NativePlatformPublishing.kt
```diff
@@ -61,8 +61,6 @@ open class NativePlatformPublishSnapshot(releaseType: ReleaseType, uploadTasks:
     params {
         param("ARTIFACTORY_USERNAME", releaseType.username)
         password("ARTIFACTORY_PASSWORD", releaseType.password, display = ParameterDisplay.HIDDEN)
-        password("pgpSigningKey", "credentialsJSON:ae6de222-f77d-4c71-b94e-fac6da485143", display = ParameterDisplay.HIDDEN)
```
wolfs, line 64: The downside of moving the parameter further up is that all projects in Gradle Native can access is. Given how we handle the key elsewhere, I suppose that is fine.

wolfs (approved):

PR #333, opened 2023-11-14
Add flag to file-events build to build without outline atomics

bratkartoffel:
This should fix gradle/gradle#24875, as no __getauxval calls will be generated by gcc.
I couldn't find a way to test this, so I'd really appreciate if someone could build a nightly version for me to test.

### DIFF
--- a/file-events/build.gradle
+++ b/file-events/build.gradle
@@ -33,6 +33,7 @@ model {
                     cppCompiler.args "-pthread"                 // Force nicer threading
                     cppCompiler.args "-pedantic"                // Disable non-standard things
                     cppCompiler.args "--std=c++11"              // Enable C++11
+                    cppCompiler.args "-mno-outline-atomics"     // Disable __getauxval calls, fixes gradle/gradle#24875 on aarch64
                     cppCompiler.args "-Wall"                    // All warnings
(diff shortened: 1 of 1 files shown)


lptr:
According to this answer on SO: https://stackoverflow.com/a/65240748/104894, this flag is only supported on GCC 9.4. We are building with GCC 5.x on Linux, so this doesn't seem to work. :/

bratkartoffel:
Shouldn't you just upgrade your GCC Installation?

<https://gcc.gnu.org/releases.html>

Latest 5.x (5.5) is from oct 2017, 9.4 is from jun 2021. so 9.4 should be OK?


Am 23. Februar 2024 08:13:06 UTC schrieb "Lóránt Pintér" ***@***.***>:

lptr:
> Shouldn't you just upgrade your GCC Installation? <[gcc.gnu.org/releases.html](https://gcc.gnu.org/releases.html)> Latest 5.x (5.5) is from oct 2017, 9.4 is from jun 2021. so 9.4 should be OK?

I thought this was not possible if we wanted to support Ubuntu 16.04, but it seems that we technically could, though it's not trivial. There is a different option, though, that I'm more keen to spend time on when we can, see: https://github.com/gradle/gradle/issues/24875#issuecomment-1964165475

PR #336, opened 2024-03-19
CI: make sure CI always picks CentOS Stream 9

cobexer:


cobexer:
@lptr could you look into the test failures on Windows and macOS so I can test on CentOS 9?

lptr:
I don't even understand how they can be so flaky. And the Windows test I don't even see failing in the linked build. :/

Anyway, I won't have time anytime soon to look at this. I'll circle back when I can take a better look.

lptr:
:tada:

PR #341, opened 2024-08-23
Remove file-events subproject

lptr:
This is now handled as a separate library:

- https://github.com/gradle/gradle-fileevents


.teamcity/Agent.kt
```diff
@@ -16,7 +16,7 @@
 
 enum class Agent(val os: Os, val architecture: Architecture) {
     UbuntuAmd64(os = Os.Ubuntu16, architecture = Architecture.Amd64),
-    UbuntuAarch64(os = Os.Ubuntu16, architecture = Architecture.Aarch64),
+    UbuntuAarch64(os = Os.Ubuntu24, architecture = Architecture.Aarch64),
```
lptr, line 19: We need to revert this to Ubuntu16 until we can do cross-compilation with Zig.

.teamcity/Os.kt
```diff
@@ -27,6 +27,13 @@ interface Os {
         }
     }
 
+    object Ubuntu24 : Linux(Ncurses.Ncurses6) {
+        override fun Requirements.additionalRequirements() {
+            contains(osDistributionNameParameter, "ubuntu")
+            contains(osDistributionVersionParameter, "24")
+        }
+    }
+
```
lptr, line 36: :x: We don't need this.

asodja (approved):
LGTM

PR #344, opened 2024-10-21
Fix architecture detection on macOS

adammurdoch:


ghale (approved):
LGTM

native-platform/src/main/cpp/posix.cpp
```diff
@@ -52,10 +53,38 @@ Java_net_rubygrapefruit_platform_internal_jni_NativeLibraryFunctions_getSystemIn
...
     jfieldID hostnameField = env->GetFieldID(infoClass, "hostname", "Ljava/lang/String;");
     env->SetObjectField(info, hostnameField, char_to_java(env, machine_info.nodename, result));
+
+    jfieldID machineArchitectureField = env->GetFieldID(infoClass, "machineArchitecture", "Ljava/lang/String;");
+#ifndef __APPLE__
+    env->SetObjectField(info, machineArchitectureField, char_to_java(env, machine_info.machine, result));
+#else
+    // On macOS, uname() reports the architecture of the current binary.
+    // Instead, use a macOS specific sysctl() to query the CPU name, which can be mapped to the architecture
+    int mib[5];
```
jvandort, line 65: why 5? I'm having trouble finding official documentation for "machdep.cpu.brand_string"

jvandort (approved):
This seems reasonable

native-platform/src/main/cpp/posix.cpp
adammurdoch: It's 5 because that's what it was in the example I copied it from. 4 also works. 3 and lower do not.

It's kind of documented on the sysctl man page.

PR #353, opened 2025-01-30
Require explicit initialization of native integration

lptr:
Previously we'd implicitly initialize the native integration, causing binaries to be extracted in the system temp directory. This is a security vulnerability and thus is not allowed anymore. Instead we now throw if native integrations are queried without initialization.

### DIFF
--- a/native-platform/src/main/java/net/rubygrapefruit/platform/Native.java
+++ b/native-platform/src/main/java/net/rubygrapefruit/platform/Native.java
@@ -40,8 +40,7 @@ public class Native {
     /**
      * Initialises the native integration, if not already initialized.
      *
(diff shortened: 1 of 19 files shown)


ljacomet (changes requested):
`net.rubygrapefruit.platform.test.Main` should be adapted as well to make the `cache-dir` CLI option mandatory.

lptr:
> `net.rubygrapefruit.platform.test.Main` should be adapted as well to make the `cache-dir` CLI option mandatory.

Why? This is a test app to start with, and sans a `cache-dir` option we now create the cache dir in the working directory (so not in system temp).

ljacomet (approved):
Looks like I missed the changes to `Main`, LGTM!

PR #356, opened 2025-05-12
Fix macOS Intel architecture detection

MSDarwish2000:
Newer Intel CPUs Display Name doesn't start with "Intel". For example, it can look like "13th Gen Intel(R) Core(TM) i5-13600K". While newer Intel processors are not officially supported by macOS, those can be present in experimental setups.

A better approach would be finding a different way than depending on display name as even Apple may change their mind about how to structure it.

### DIFF
--- a/native-platform/src/main/java/net/rubygrapefruit/platform/internal/MutableSystemInfo.java
+++ b/native-platform/src/main/java/net/rubygrapefruit/platform/internal/MutableSystemInfo.java
@@ -39,8 +39,7 @@ public class MutableSystemInfo implements SystemInfo {
             // On macOS, the architecture field contains the CPU name, not the architecture
             if (machineArchitecture.startsWith("Apple")) {
                 return "arm64";
-            }
-            if (machineArchitecture.startsWith("Intel")) {
+            } else if (machineArchitecture.contains("Intel") || machineArchitecture.startsWith("AMD")) {
                 return "x86_64";
             }
         }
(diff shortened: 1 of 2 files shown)


ov7a (changes requested):
Thank you for your contribution!

The [DCO](https://probot.github.io/apps/dco/) check is failing, please look into that as we can't accept the PR otherwise.

***

Can you provide more details about why you should change `startsWith` to `contains` for Apple? Example strings can help. Covering some sample strings with unit tests would work too.

Can you also provide the string with CPU name from the output for `sysctl machdep` for the hardware in question?

MSDarwish2000:
> The [DCO](https://probot.github.io/apps/dco/) check is failing, please look into that as we can't accept the PR otherwise.

I have fixed it. Thank you for the clarification.

> Can you provide more details about why you should change `startsWith` to `contains` for Apple? 

It was more of futureproofing if Apple decided to change it similar to Intel. On the latest commit, I reverted it.

> Example strings can help. Covering some sample strings with unit tests would work too.
>
> Can you also provide the string with CPU name from the output for `sysctl machdep` for the hardware in question?

I have written a unit test for this. Most values are retrieved by searching online on CPU databases. My output is:

```
% sysctl -n machdep.cpu.brand_string
13th Gen Intel(R) Core(TM) i5-13600K
```

ov7a (approved):

PR #364, opened 2025-08-27
fix upload jni task

reinsch82:
Fixing issues related to signing and uploading JNI jars


buildSrc/src/main/java/gradlebuild/NativePlatformComponentPlugin.java
```diff
@@ -103,7 +115,8 @@ public void execute(Task task) {
         // This allows using dependency substitution for the root project.
         DependencyHandler dependencies = project.getDependencies();
         project.getDependencies().add("testImplementation", project.getDependencies().project(ImmutableMap.of("path", project.getPath())));
-        dependencies.add("testImplementation", "org.spockframework:spock-core:1.3-groovy-2.5");
+        dependencies.add("testImplementation", "org.spockframework:spock-core:2.3-groovy-3.0");
+        dependencies.add("testImplementation", "junit:junit:4.12");
```
cobexer, line 119: JUnit 4.12 is from Dec 04, 2014 and has a CVE, could we at least use JUnit 4.13.2 from 2021?

gradle/verification-metadata.xml
cobexer, line 1: Could you attempt to regenerate this without the <ignored-keys> and <components> in place?

Sometimes key server rate limit / block you and that leads to many key misses ... maybe that hit you here?

cobexer (changes requested):
I see no tests executing on CI any more, PTAL why that no longer happens

PR #378, opened 2026-01-20
Compile to Java 8 for native platform

asodja:
Since we use it in gradle/gradle for in example [process-memory-services](https://github.com/gradle/gradle/blob/4cb26df97ce64a9ca2804342a9646a615527bff8/platforms/core-runtime/process-memory-services/build.gradle.kts#L26) that is [used for workers](https://github.com/gradle/gradle/blob/4cb26df97ce64a9ca2804342a9646a615527bff8/platforms/core-runtime/process-memory-services/build.gradle.kts#L7-L11) that can run on Java 8.

### DIFF
--- a/test-app/build.gradle
+++ b/test-app/build.gradle
@@ -41,10 +41,11 @@ publishing {
 
 // Download and unpack the application to allow it to be tested
 tasks.register('download') {
+    def downloadDir = project.layout.buildDirectory.dir("download")
     doLast {
         copy {
             from(configurations.download.files.collect { zipTree(it) })
-            into("$buildDir/download")
+            into(downloadDir)
         }
     }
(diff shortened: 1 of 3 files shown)


test-app/build.gradle
cobexer: 💅 We should make sure to run this app with Java 8.

Please add some toolchain configuration!

cobexer (commented):
LGTM, making sure the test app runs with Java 8 would be great!

test-app/build.gradle
asodja: Good point

cobexer (approved):

ov7a (approved):

eskatos (approved):

